PR #208, opened 2026-02-26
feat(tui): add file-watcher subscription for standalone mf tui command

H31MDALLR:
## Summary
- Add `file_subscription.clj` that tail-follows `~/.miniforge/events/*.edn` files using `RandomAccessFile` with position tracking, enabling cross-process TUI monitoring
- Add `start-standalone-tui!` entry point in `interface.clj` wired to file-based subscription instead of in-memory event stream
- Update `monitoring/tui-cmd` and add dedicated `tui.clj` CLI command to launch standalone TUI via `requiring-resolve`
- Make `translate-event` public in `subscription.clj` for reuse by both in-memory and file-based subscription paths
- Fix pre-existing unresolved `clojure.string` namespace in `monitoring.clj`

## Test plan
- [ ] Verify `mf tui` launches and displays the TUI interface
- [ ] Run a workflow with `mf run`, confirm TUI picks up events from `~/.miniforge/events/`
- [ ] Start TUI before any workflow runs, confirm it discovers new event files when a workflow starts
- [ ] Test `send-command!` writes correct `.edn` to `~/.miniforge/commands/<id>/`
- [ ] Verify cleanup function stops the polling thread
- [ ] Confirm existing `start-tui!` (in-memory mode) still works unchanged

### DIFF
--- a/components/tui-views/src/ai/miniforge/tui_views/file_subscription.clj
+++ b/components/tui-views/src/ai/miniforge/tui_views/file_subscription.clj
@@ -1,0 +1,182 @@
+;; Title: Miniforge.ai
+;; Subtitle: An agentic SDLC / fleet-control platform
+;; Author: Christopher Lester
+;; Line: Founder, Miniforge.ai (project)
+;; Copyright 2025-2026 Christopher Lester (christopher@miniforge.ai)
+;;
+;; Licensed under the Apache License, Version 2.0 (the "License");
+;; you may not use this file except in compliance with the License.
+;; You may obtain a copy of the License at
+;;
+;;     http://www.apache.org/licenses/LICENSE-2.0
+;;
+;; Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


components/tui-views/src/ai/miniforge/tui_views/file_subscription.clj
```diff
@@ -0,0 +1,182 @@
...
+  (let [poll-ms   (get opts :poll-ms 500)
+        scan-ms   (get opts :scan-ms 2000)
+        dir       (events-dir)
+        ;; Map of file-path -> position-atom for tracked files
+        tracked   (atom {})
+        running?  (atom true)
+
+        ;; Track a file: read all existing lines (hydrate), store position
+        track-file!
+        (fn [^java.io.File f]
```
H31MDALLR: Extract these embedded anonymous functions out to their own function.

PR #211, opened 2026-02-27
feat(workflow): add chain lifecycle event emission

H31MDALLR:
## Summary

- Add 6 chain event constructors to `event-stream` component (`chain-started`, `chain-step-started`, `chain-step-completed`, `chain-step-failed`, `chain-completed`, `chain-failed`)
- Add `emit!` helper to `chain.clj` using `requiring-resolve` for cross-component calls
- Modify `run-chain` to emit lifecycle events at each stage (start, step start/complete/fail, chain complete/fail)
- Events are conditional on `:event-stream` presence in opts — backward compatible

## Test plan

- [x] Success path: 2-step chain emits `started, step-started, step-completed, step-started, step-completed, completed`
- [x] Failure path: failing step emits `started, step-started, step-failed, failed`
- [x] No event-stream: chain runs silently without errors
- [ ] Note: pre-existing `chain_test.clj` failures exist on main (`:chain/input.task` keyword namespace parsing issue) — unrelated to this PR

### DIFF
--- a/components/event-stream/src/ai/miniforge/event_stream/core.clj
+++ b/components/event-stream/src/ai/miniforge/event_stream/core.clj
@@ -387,6 +387,54 @@
       (cond-> content (assoc :annotation/content content))))
 
 ;------------------------------------------------------------------------------ Layer 4
+;; Chain lifecycle events
+;; Chains are not workflow-scoped, so workflow-id is nil.
+;; Message is derived from the event-type keyword.
+
+(defn- chain-envelope
+  "Create an envelope for chain events. Chains are not workflow-scoped,
+   so workflow-id is nil. Message is derived from the event-type keyword."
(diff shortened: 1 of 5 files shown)


components/event-stream/src/ai/miniforge/event_stream/core.clj
```diff
@@ -386,6 +386,49 @@
              :annotation/type annotation-type)
       (cond-> content (assoc :annotation/content content))))
 
+;------------------------------------------------------------------------------ Layer 4
+;; Chain lifecycle events
+
+(defn chain-started [stream chain-id step-count]
```
H31MDALLR, line 400: So why are we carrying around strings when we can just regenerate this from the key word that we're sending in? And why is the third param here always nil?

PR #213, opened 2026-02-27
feat(tui): chain lifecycle event handlers

H31MDALLR:
## Summary
- Wire all 6 chain events (`:chain/started`, `step-started`, `step-completed`, `step-failed`, `completed`, `failed`) through the full TUI event pipeline
- Add message constructors (`msg.clj`), event translation (`subscription.clj`), model state tracking (`model.clj`), update handlers (`events.clj`), and root dispatch (`update.clj`)
- Track active chain with step progress via `:active-chain` in model — set on chain start, updated on step transitions, cleared on completion
- 12 tests covering translation and handler model state transitions

## Test plan
- [x] All 12 chain event tests pass (37 assertions)
- [x] Full `bb test :tui-views` passes (0 failures)
- [x] Pre-commit hooks pass (lint, format, test, GraalVM)

### DIFF
--- a/components/tui-views/src/ai/miniforge/tui_views/update/events.clj
+++ b/components/tui-views/src/ai/miniforge/tui_views/update/events.clj
@@ -140,6 +140,73 @@
         (apply-gate-result idx workflow-id gate passed? payload)
         with-timestamp)))
 
+;------------------------------------------------------------------------------ Layer 2b
+;; Chain event handlers
+
+(defn handle-chain-started
+  "Set active chain in model when a chain begins execution."
+  [model {:keys [chain-id step-count]}]
+  (-> model
+      (assoc :active-chain {:chain-id chain-id
+                            :step-count step-count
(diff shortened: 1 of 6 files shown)


components/tui-views/src/ai/miniforge/tui_views/update/events.clj
```diff
@@ -140,6 +140,73 @@
...
+;------------------------------------------------------------------------------ Layer 2b
+;; Chain event handlers
+
+(defn handle-chain-started
+  "Set active chain in model when a chain begins execution."
+  [model {:keys [chain-id step-count]}]
+  (-> model
+      (assoc :active-chain {:chain-id chain-id
+                            :step-count step-count
+                            :current-step nil
```
H31MDALLR, line 152: OK, so unless that is in the schema, which it may be, you can actually just leave out current step, and it'll still be a nil value on access.

PR #222, opened 2026-02-27
fix(event-stream+workflow): flush event sinks and wire phase retry recovery

H31MDALLR:
## Summary
- Event file sink was not calling `.flush()` after writes, causing events to sit in `BufferedWriter` memory instead of reaching disk. TUI and web dashboard depend on the file stream for live monitoring — this caused the monitoring views to stall mid-workflow.
- Pipeline executor did not handle `:retrying` status from phase error handlers, causing test/pre-commit failures to skip to the next phase instead of re-running the current one. Also wires `:redirect-to` for failed phases to jump back to earlier phases (e.g., verify failure -> implement).

## Test plan
- [x] Full poly test suite passes (0 failures, 0 errors)
- [x] GraalVM compatibility tests pass
- [ ] Manually verify event file updates in real-time during `mf run`
- [ ] Trigger a verify phase failure and confirm retry behavior

### DIFF
--- a/components/event-stream/src/ai/miniforge/event_stream/sinks.clj
+++ b/components/event-stream/src/ai/miniforge/event_stream/sinks.clj
@@ -61,7 +61,8 @@
         (let [file-path (event-file-path workflow-id)]
           (with-open [writer (io/writer file-path :append true)]
             (.write writer (pr-str event))
-            (.write writer "\n")))
+            (.write writer "\n")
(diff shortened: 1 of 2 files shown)


H31MDALLR:
Closing — superseded by #223 which includes both the event sink flush fix and the phase retry/redirect logic (with additional redirect cycle protection to prevent infinite loops).

PR #224, opened 2026-03-01
refactor: data-driven MCP artifact server + generic session handling

H31MDALLR:
## Summary

- Refactor MCP server from bespoke handlers to a table-driven `tool-registry`
- Add `submit_test_artifact` and `submit_release_artifact` MCP tools
- Make `artifact_session.clj` handle any artifact type generically (pattern-based UUID/instant detection)
- Wire implementer and planner agents to use MCP artifact sessions
- Add `mcp-config` passthrough to LLM CLI backend
- Add clj-kondo hook for `with-artifact-session` macro

## Test plan

- [x] All existing tests pass (lint, unit, graalvm)
- [ ] Verify `submit_test_artifact` writes correct EDN via JSON-RPC
- [ ] Verify `submit_release_artifact` writes correct EDN via JSON-RPC
- [ ] Verify `parse-uuid-strings` handles `:test/id`, `:release/id`, etc.

See `docs/pull-requests/2026-02-28-refactor-data-driven-mcp-artifact-server.md`

### DIFF
--- a/.clj-kondo/hooks/artifact_session.clj
+++ b/.clj-kondo/hooks/artifact_session.clj
@@ -1,0 +1,15 @@
+(ns hooks.artifact-session
+  (:require [clj-kondo.hooks-api :as api]))
+
+(defn with-artifact-session
+  "Expands (with-artifact-session [session] body...) to (let [session nil] body...)
+   so clj-kondo can resolve the binding."
+  [{:keys [node]}]
+  (let [children (rest (:children node))
+        binding-vec (first children)
+        session-sym (first (:children binding-vec))
+        body (rest children)]
+    {:node (api/list-node
+            (list* (api/token-node 'let)
+                   (api/vector-node [session-sym (api/token-node nil)])
+                   body))}))
--- a/components/agent/resources/prompts/planner.edn
+++ b/components/agent/resources/prompts/planner.edn
@@ -95,7 +95,8 @@ Call `submit_plan` with:
 
 Do NOT output raw EDN text when the MCP tool is available.
 
-If the tool is NOT available, fall back to the EDN output format described above.
+The MCP tool is always available in this environment. If you encounter an error calling it,
+report the error rather than falling back to raw text output.
 
 Remember: A good plan enables parallel execution where possible while respecting
 necessary dependencies. Optimize for clarity and testability."}
--- a/components/agent/resources/prompts/planner.edn
+++ b/components/agent/resources/prompts/planner.edn
@@ -83,5 +83,19 @@ You must output a structured plan as a Clojure map with:
 9. **Review Points**: Include :review tasks at natural checkpoints, especially after
    significant implementation milestones.
 
+## Preferred Output Method
+
+If the `submit_plan` MCP tool is available, you MUST use it to submit your plan.
+Call `submit_plan` with:
+- `name`: short descriptive name for the plan
+- `tasks`: array of task objects with `description`, `type` (\"implement\", \"test\", \"review\", \"design\", \"deploy\", \"configure\"), and optionally `dependencies` (array of task indexes), `acceptance_criteria`, `estimated_effort`
+- `complexity`: \"low\", \"medium\", or \"high\" (optional)
+- `risks`: array of risk strings (optional)
+- `assumptions`: array of assumption strings (optional)
+
+Do NOT output raw EDN text when the MCP tool is available.
+
+If the tool is NOT available, fall back to the EDN output format described above.
+
 Remember: A good plan enables parallel execution where possible while respecting
 necessary dependencies. Optimize for clarity and testability."}
--- a/components/agent/src/ai/miniforge/agent/artifact_session.clj
+++ b/components/agent/src/ai/miniforge/agent/artifact_session.clj
@@ -1,0 +1,197 @@
+(ns ai.miniforge.agent.artifact-session
+  "MCP artifact session management.
+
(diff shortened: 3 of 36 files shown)


.clj-kondo/hooks/artifact_session.clj
```diff
@@ -0,0 +1,15 @@
+(ns hooks.artifact-session
+  (:require [clj-kondo.hooks-api :as api]))
+
+(defn with-artifact-session
+  "Expands (with-artifact-session [session] body...) to (let [session nil] body...)
```
H31MDALLR, line 5: Do we have this function only for Lint?

components/agent/resources/prompts/planner.edn
```diff
@@ -83,5 +83,19 @@ You must output a structured plan as a Clojure map with:
...
+Call `submit_plan` with:
+- `name`: short descriptive name for the plan
+- `tasks`: array of task objects with `description`, `type` (\"implement\", \"test\", \"review\", \"design\", \"deploy\", \"configure\"), and optionally `dependencies` (array of task indexes), `acceptance_criteria`, `estimated_effort`
+- `complexity`: \"low\", \"medium\", or \"high\" (optional)
+- `risks`: array of risk strings (optional)
+- `assumptions`: array of assumption strings (optional)
+
+Do NOT output raw EDN text when the MCP tool is available.
+
+If the tool is NOT available, fall back to the EDN output format described above.
```
H31MDALLR: Is having a fallback wise? This feels like we will silently roll over into a generated EDN that may or may not adhere to our specs. Should we instead detect that the MCP tool is not installed and install them?

components/agent/src/ai/miniforge/agent/artifact_session.clj
```diff
@@ -0,0 +1,197 @@
...
+
+   Returns:
+   - {:dir <path>, :mcp-config-path <path>, :artifact-path <path>}"
+  []
+  (let [dir (str (Files/createTempDirectory
+                  "miniforge-artifact-"
+                  (into-array FileAttribute [])))
+        config-path (str dir "/mcp-config.json")
+        artifact-path (str dir "/artifact.edn")]
+    {:dir dir
```
H31MDALLR, line 67: Make sure things like this have a schema, a Mali schema that we can validate against.

components/agent/src/ai/miniforge/agent/planner.clj
```diff
@@ -272,37 +273,46 @@
...
+                               (parse-plan-response content)
+                               (make-fallback-plan spec-text))]
+                  (when artifact
+                    (log/info logger :planner :planner/mcp-artifact-received
+                              {:data {:task-count (count (:plan/tasks artifact))}}))
+                  ;; Ensure all tasks have proper UUIDs
+                  (let [plan-with-ids (-> plan
+                                          (update :plan/id #(or % (random-uuid)))
+                                          (update :plan/tasks
+                                                  (fn [tasks]
```
H31MDALLR: Consider extracting this since we have an anon function inside of an anon function.

components/agent/src/ai/miniforge/agent/planner.clj
```diff
@@ -272,37 +273,46 @@
...
+                                                    (mapv (fn [t]
+                                                            (update t :task/id #(or % (random-uuid))))
+                                                          tasks)))
+                                          (assoc :plan/created-at (java.util.Date.)))]
+                    {:status :success
```
H31MDALLR: Success and failure should use our patterns we have established with response successes and anomalies.

.clj-kondo/hooks/artifact_session.clj
H31MDALLR: Yes — this is a clj-kondo analysis hook only. The `with-artifact-session` macro uses a binding form `[session]` that clj-kondo can't resolve without a hint. The hook expands it to a `let` so the `session` symbol is resolvable during linting. No runtime effect.

components/agent/resources/prompts/planner.edn
H31MDALLR: Good point. Removed the fallback instruction from both planner.edn and implementer.edn. The MCP tool is always provisioned by the artifact session, so it will always be available. The prompt now says: "The MCP tool is always available in this environment. If you encounter an error calling it, report the error rather than falling back to raw text output."

components/agent/src/ai/miniforge/agent/artifact_session.clj
H31MDALLR: Added a `Session` Malli schema and `validate-session` function to artifact_session.clj. The schema validates that `:dir`, `:mcp-config-path`, and `:artifact-path` are all non-empty strings.

components/agent/src/ai/miniforge/agent/planner.clj
H31MDALLR: Extracted into two named functions: `ensure-task-ids` (maps over tasks to fill missing UUIDs) and `finalize-plan` (composes ID/task-ID/timestamp fixup). The invoke-fn now just calls `(finalize-plan plan)`.

components/agent/src/ai/miniforge/agent/planner.clj
H31MDALLR: Switched to `response/success` and `response/error` from the response interface. Success path now uses `(response/success plan-final {:tokens tokens :metrics {...}})` and error path uses `(response/error error-msg {:output fallback})`.

PR #228, opened 2026-03-01
test: MCP artifact server + session tests + parallel pre-commit

H31MDALLR:
## Summary

- Zero-to-full test coverage for MCP artifact server (50 assertions via subprocess JSON-RPC) and artifact session (10 unit tests, 45 assertions)
- Integration tests for full round-trip: create-session → spawn MCP server → submit artifact → read-artifact
- Surfaced **two latent hang vectors** that explain the dogfooding stall (details in PR doc)
- Replaced serial `poly test` in pre-commit with parallel pmap-based runner: queries poly for changed bricks since `main`, runs only their tests in parallel — **pre-commit down from 5+ minutes to ~13 seconds**

## Root Cause Findings

1. **`when-let` nil swallow** (mcp-artifact-server.bb:409) — if `dispatch` returns `nil` for a request, no JSON-RPC response is sent, client deadlocks
2. **Stderr pipe buffer deadlock** (llm_client.clj:390) — `:err :string` buffers stderr; if MCP server fills the 64KB pipe buffer with diagnostics, three-process deadlock

## Test plan

- [x] `bb test:mcp` — 50/50 passing (MCP server handler tests)
- [x] `bb test` — 137 tests, 824 assertions, 0 failures (parallel, ~13s)
- [x] `bb pre-commit` — full pass in ~13s (lint + fmt + parallel tests + graalvm)
- [ ] `bb test:integration` — includes MCP round-trip integration test

### DIFF
--- a/bb.edn
+++ b/bb.edn
@@ -105,7 +105,69 @@
   ;; ──────────────────────────────────────────────────────────────────────────
 
   test
-  {:doc  "Run unit (brick) tests via Polylith"
+  {:doc  "Run changed brick tests in parallel (since main)"
+   :task (do
+           (println "🧪 Detecting changed bricks since main...")
+           ;; Query poly for changed components + bases since main
+           (let [{:keys [out exit]} (p/sh {:out :string}
+                                          "poly" "ws" "get:changes:changed-components" "since:main")
+                 changed-bases (:out (p/sh {:out :string}
+                                           "poly" "ws" "get:changes:changed-bases" "since:main"))
+                 ;; Parse poly EDN output (e.g. ["agent" "llm"])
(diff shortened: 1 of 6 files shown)


bb.edn
```diff
@@ -105,7 +105,69 @@
   ;; ──────────────────────────────────────────────────────────────────────────
 
   test
-  {:doc  "Run unit (brick) tests via Polylith"
+  {:doc  "Run changed brick tests in parallel (since main)"
+   :task (do
```
H31MDALLR: We should have this kind of code inside of CLJ (or .bb) files and not inside of BB this way we can actually test them and bb.edn does not grow to extreme size.

PR #230, opened 2026-03-02
feat: explore phase, evidence bundles, config-driven file loading

H31MDALLR:
## Summary
- Adds a deterministic **explore phase** to the standard SDLC pipeline that loads existing files into context before planning, giving the planner visibility into what already exists
- When the planner detects work is **already satisfied**, it now returns a **structured evidence bundle** (`[{:requirement :satisfied-by :proof}]`) proving each spec requirement is met
- Makes file context limits (`max-files`, `max-lines-per-file`) **config-driven** instead of hardcoded constants
- Expands directory scope paths via glob so the explore phase actually finds files
- Skips gate validation and treats `:already-satisfied` as success in the execution engine
- Fixes `runner_test` to not depend on LLM infrastructure

## Test plan
- [x] All 14 test namespaces pass (37+ assertions per namespace, 0 failures)
- [x] GraalVM compatibility tests pass
- [x] Pre-commit hooks (lint, format, test) all pass
- [ ] Run `bb miniforge run work/pr-lifecycle-tests.spec.edn` to verify already-satisfied detection with evidence bundle


components/agent/src/ai/miniforge/agent/planner.clj
```diff
@@ -282,54 +285,78 @@
...
-                                               :complexity (:plan/estimated-complexity plan-final)
-                                               :tokens tokens}}))
+                                :streaming? (boolean on-chunk)}})
+              (if (llm/success? response)
+                (let [content (llm/get-content response)
+                      plan (or (parse-plan-response content)
+                               (make-fallback-plan spec-text))]
+                  ;; Check for already-satisfied response
+                  (if (= :already-satisfied (:plan/status plan))
+                    {:status :already-satisfied
```
H31MDALLR: Do we not have standard response fns for success and anomalies?

components/phase/src/ai/miniforge/phase/verify.clj
H31MDALLR, line 87: This assumes we will always be testing in Clojure. Should we assume that? What about when the plan calls for a tech stack written in Rust? Still, we can probably decide that all tests will be run in Clojure if needed

H31MDALLR (commented):
Addressed both comments in 534f112:

1. **planner.clj** — Now uses `response/success` builder instead of inline map, then overrides `:status` to `:already-satisfied`.

2. **verify.clj** — Added NOTE documenting the Clojure-specific assumption in `parse-test-output`. When we support other stacks (Rust, JS), this should dispatch on `:test/framework` or `:language` from workflow config. Fine for YC MVP since we're Clojure-only.

PR #231, opened 2026-03-02
fix(plan): fail with evidence, universal DAG executor, MCP component

H31MDALLR:
## Summary

- Remove silent `make-fallback-plan` — throw with evidence when both MCP
  artifact and EDN parse fail
- Extract MCP artifact server script into proper Polylith component
  (`components/mcp-artifact-server/`)
- Wire DAG executor as universal executor for all plans after plan phase
- Detect spec-provided `:plan/tasks` and skip LLM call entirely
- Fix `artifact_session.clj`: throw on server resolution failure, log parse errors

## Test plan

- [x] `bb pre-commit` passes (380 tests, 0 failures, 0 errors + GraalVM compat)
- [ ] Run `finish-yc-mvp.spec.edn` — should skip LLM planning, route to DAG
- [ ] Verify plan failure evidence — mock LLM returning garbage, confirm exception with data
- [ ] Test MCP server component unit tests in isolation
- [ ] Verify DAG executor handles single-task plans correctly


components/mcp-artifact-server/src/ai/miniforge/mcp_artifact_server/server.clj
```diff
@@ -0,0 +1,88 @@
...
+  "Run the MCP server loop, reading JSON-RPC messages from stdin.
+
+   Arguments:
+   - artifact-dir — directory path for artifact persistence"
+  [artifact-dir]
+  (binding [*out* *err*]
+    (println "miniforge-artifact MCP server started, artifact-dir:" artifact-dir))
+  (let [reader (java.io.BufferedReader. (java.io.InputStreamReader. System/in "UTF-8"))]
+    (loop []
+      (when-let [line (.readLine reader)]
```
H31MDALLR, line 110: This looks pretty messy. There's a when, when not, one, two nested tries, and then nested ifs, and work going on in the catch. Extract, decompose this, make it idiomatic.

components/mcp-artifact-server/src/ai/miniforge/mcp_artifact_server/tools.clj
```diff
@@ -0,0 +1,335 @@
...
+  [required-params params]
+  (doseq [[param-name {:keys [check msg]}] required-params]
+    (let [v (get params param-name)]
+      (when-not (and v (check v))
+        (throw (ex-info msg {:code -32602}))))))
+
+;------------------------------------------------------------------------------ Layer 1
+;; Tool registry
+
+(def tool-registry
```
H31MDALLR: Why is the tool registry in code? This should be something we can configure via data.

components/mcp-artifact-server/src/ai/miniforge/mcp_artifact_server/tools.clj
```diff
@@ -0,0 +1,335 @@
...
+       "tests_needed"
+       {:type "boolean"
+        :description "Whether tests should be written for this code"}
+       "dependencies_added"
+       {:type "array"
+        :description "List of dependencies added (e.g. [\"lib/name {:mvn/version \\\"1.0.0\\\"}\"])"
+        :items {:type "string"}}}}}
+
+    :required-params
+    {"files"   {:check #(seq %)              :msg "files is required and must not be empty"}
```
H31MDALLR: If we are going to have functions embedded in this configuration, we probably want to extract the functions, use a keyword lookup for the functions, so that we can move all of this configuration to data. Give it some thought.

components/mcp-artifact-server/src/ai/miniforge/mcp_artifact_server/tools.clj
```diff
@@ -0,0 +1,335 @@
...
+       {:type "array"
+        :description "Assumptions made during planning"
+        :items {:type "string"}}}}}
+
+    :required-params
+    {"name"  {:check #(not (str/blank? %)) :msg "name is required"}
+     "tasks" {:check #(seq %)              :msg "tasks is required and must not be empty"}}
+
+    :build-artifact
+    (fn [params]
```
H31MDALLR: At any rate, all of these anonymous functions should be extracted. Some of them are also anonymous inside of anonymous, and those should be decomposed and extracted.

components/phase/src/ai/miniforge/phase/plan.clj
```diff
@@ -72,55 +72,75 @@
...
-        (assoc-in [:phase :agent] :planner)
-        (assoc-in [:phase :gates] gates)
-        (assoc-in [:phase :budget] budget)
-        (assoc-in [:phase :started-at] start-time)
-        (assoc-in [:phase :status] :running)
-        (assoc-in [:phase :result] result))))
+        input (get-in ctx [:execution/input])]
+
+    ;; Fast path: spec provides explicit tasks — skip LLM
+    (if-let [spec-tasks (:plan/tasks input)]
```
H31MDALLR: This is also a good candidate for decomposition into functions. The fast path is a function, the normal path is a function, et cetera.

components/workflow/resources/workflows/nimble-sdlc-v1.0.0.edn
H31MDALLR, line 4: I'm not sure that Agile was represented here, because Agile is the entire process. It's just a much faster process with a small plan, which would be a delta from the waterfall plan:
- POC
- dev implementation
- test implementation
- adjusting the major plan as you move forward

In Nimble, you do the waterfall planning up front. Then you have that set of decomposed things that you can then do Agile sprint to milestones. In Agile, you would decide whether or not you change those goals or move them to other milestones.

components/workflow/src/ai/miniforge/workflow/execution.clj
```diff
@@ -313,6 +314,50 @@
...
+
+(defn- try-dag-execution
+  "After plan phase, execute all plans via the DAG executor.
+
+   The DAG executor is the universal executor — it handles both parallel
+   and sequential plans. Returns updated context with DAG results and
+   skipped-to index, or nil if this isn't the plan phase or no plan was produced."
+  [ctx-processed phase-name phase-result pipeline
+   transition-to-completed-fn transition-to-failed-fn]
+  (when (= :plan phase-name)
```
H31MDALLR: When, when, when, if, if. That's five levels of nesting for this. Refactor into a more idiomatic pipeline-based code. We should remove as many these state variables created as a result of the conditions as we can . They should be declarative rather than conditional because conditionals add complexity with each branch .. and this is 2^6 branch possibilities.

components/mcp-artifact-server/src/ai/miniforge/mcp_artifact_server/tools.clj
```diff
@@ -0,0 +1,376 @@
...
+   Keeping this separate from the registry enables the registry to be pure data."
+  {"submit_code_artifact"    build-code-artifact
+   "submit_plan"             build-plan-artifact
+   "submit_test_artifact"    build-test-artifact
+   "submit_release_artifact" build-release-artifact})
+
+;------------------------------------------------------------------------------ Layer 1
+;; Tool registry — pure data, no embedded functions
+
+(def tool-registry
```
H31MDALLR: Now that this is pure data .. it should be configuration. We need to think through how we add tools in an extensible way.

PR #248, opened 2026-03-05
Fix metrics tracking: token counts, cost, and duration

H31MDALLR:
## Summary

- **Parse Claude CLI `"result"` event** for token usage — was previously ignored (returned `nil`), causing all metrics to propagate as zeros through the pipeline
- **Thread usage data through streaming pipeline** via `accumulated-usage` atom in `stream-with-parser` → `handle-streaming` → `streaming-success-response`
- **Add `:tokens` key** to all success response paths (streaming, non-streaming CLI, OpenAI HTTP) for consistent downstream access
- **Estimate `cost-usd`** at the implement phase level using a blended rate ($15/1M tokens)
- **Aggregate `cost-usd` + `duration-ms`** alongside tokens in DAG orchestrator's `aggregate-results` and `dag-execution-result`

## Root cause

Claude CLI's `--output-format stream-json` emits usage in a final `{"type": "result", "result": {"usage": {...}}}` event. `parse-claude-stream-line` had a catch-all `nil` for `"result"` type (line 49), so usage was never captured. This propagated zeros through: `streaming-success-response` → `implement.clj` metrics → `dag_orchestrator` aggregation → display output showing `"Tokens: 0 | Cost: $0.0000 | Duration: 0ms"`.

## Files changed

| File | What |
|------|------|
| `components/llm/.../llm_client.clj` | Parse `"result"` event, thread `accumulated-usage` atom through streaming, add `:tokens` to all response paths |
| `components/phase/.../implement.clj` | Compute `cost-usd` from token count, merge into execution metrics |
| `components/workflow/.../dag_orchestrator.clj` | Aggregate `cost-usd` + `duration-ms`, pass full metrics map to result constructor |

## Test plan

- [x] All 601 existing tests pass (0 failures, 0 errors)
- [ ] Run a dogfood spec and verify non-zero tokens/cost/duration in output
- [ ] Verify OpenAI HTTP path still reports tokens (existing tests cover this)

### DIFF
--- a/components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
+++ b/components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
@@ -228,8 +228,8 @@
            :content (get-in body [:choices 0 :message :content] "")
            :usage {:input-tokens (get-in body [:usage :prompt_tokens])
                    :output-tokens (get-in body [:usage :completion_tokens])}
-           :tokens (+ (or (get-in body [:usage :prompt_tokens]) 0)
-                      (or (get-in body [:usage :completion_tokens]) 0))}
+           :tokens (+ (get-in body [:usage :prompt_tokens] 0)
+                      (get-in body [:usage :completion_tokens] 0))}
           {:success false
            :anomaly (response/make-anomaly
                      :anomalies/unavailable
--- a/components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
+++ b/components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
@@ -26,10 +26,12 @@
   ([content]
    (llm-success content nil))
   ([content {:keys [usage exit-code]}]
-   (cond-> {:success true
-            :content content
-            :usage (or usage {:input-tokens nil :output-tokens nil})}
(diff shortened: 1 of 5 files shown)


components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
```diff
@@ -221,7 +227,9 @@
           {:success true
            :content (get-in body [:choices 0 :message :content] "")
            :usage {:input-tokens (get-in body [:usage :prompt_tokens])
-                   :output-tokens (get-in body [:usage :completion_tokens])}}
+                   :output-tokens (get-in body [:usage :completion_tokens])}
+           :tokens (+ (or (get-in body [:usage :prompt_tokens]) 0)
```
H31MDALLR: Use the default for get-in instead of or.

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
```diff
@@ -221,7 +227,9 @@
           {:success true
```
H31MDALLR: We have success / anomaly patterns. They should be used here. I see this component does not do so.

PR #254, opened 2026-03-06
Remove silent fallback stubs from all agents

H31MDALLR:
## Summary
- Remove `make-fallback-*` functions from implementer, releaser, tester, and release-executor metadata
- All agents now fail explicitly with `:status :error` when LLM is unavailable or response parsing fails
- Fallback stubs were masking real failures, preventing retry/repair from working, and short-circuiting checkpoint resume
- Add `:release-meta` opts support in `execute-release-phase` so tests (and callers with pre-computed metadata) can bypass the releaser agent

## Test plan
- [x] All 484 core tests pass (0 failures, 0 errors)
- [x] Agent invoke/cycle tests updated to expect `:error` without LLM backend
- [x] Sandbox tests updated with pre-built release metadata
- [ ] Verify dogfood run fails explicitly instead of creating stub PRs

### DIFF
--- a/components/agent/src/ai/miniforge/agent/implementer.clj
+++ b/components/agent/src/ai/miniforge/agent/implementer.clj
@@ -5,6 +5,7 @@
    [ai.miniforge.agent.artifact-session :as artifact-session]
    [ai.miniforge.agent.prompts :as prompts]
    [ai.miniforge.agent.specialized :as specialized]
+   [ai.miniforge.response.interface :as response]
    [ai.miniforge.schema.interface :as schema]
    [ai.miniforge.logging.interface :as log]
    [ai.miniforge.llm.interface :as llm]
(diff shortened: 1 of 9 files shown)


components/agent/src/ai/miniforge/agent/implementer.clj
```diff
@@ -322,47 +309,40 @@
...
+                                         :code/files files
+                                         :code/tests-needed? true
+                                         :code/summary "Implementation from code blocks"
+                                         :code/created-at (java.util.Date.)}))]
+                      (let [code-with-meta (-> code
+                                               (update :code/id #(or % (random-uuid)))
+                                               (assoc :code/language (extract-language (:code/files code) context))
+                                               (assoc :code/created-at (java.util.Date.)))
+                            lang (:code/language code-with-meta)]
+                        {:status :success
```
H31MDALLR: We should be using our success and anomaly patterns here.

components/agent/src/ai/miniforge/agent/tester.clj
```diff
@@ -355,26 +319,19 @@
...
-                    framework (extract-test-framework (:test/files tests) context)]
-                {:status :success
-                 :output (assoc tests :test/framework framework)
-                 :artifact (assoc tests :test/framework framework)
-                 :metrics {:test-files (count (:test/files tests))
-                           :test-type (:test/type tests)
-                           :assertions 1
-                           :cases 1}})))))
+                    ;; LLM returned content but no parseable tests — fail explicitly
+                    {:status :error
```
H31MDALLR: Again, we have anomaly patterns that we should be using along with helpers.

PR #269, opened 2026-03-07
Fix dogfood Run 7: duration-ms always 0, missing terminal events, artifact error severity

H31MDALLR:
## Summary

Fixes 3 bugs discovered during dogfood Run 7 of `finish-yc-mvp.spec.edn`:

- **Duration-ms always 0**: All phases reported `"Duration: 0ms"` because `publish-phase-completed!` tried `[:metrics :duration-ms]` (agent's `{:duration-ms 0}` fallback) before `(:duration-ms result)` (real wall-clock time). Fixed extraction order in runner + overrode agent metrics with real duration in all 4 phase leave functions (plan, implement, verify, review).

- **Terminal events silently dropped**: `publish-workflow-completed!` used `when-let` around `requiring-resolve` — when the constructor var wasn't resolvable, it silently did nothing instead of falling through to the ad-hoc event fallback. Changed to `if-let` so the fallback always fires.

- **Artifact "WARN" was actually fatal**: `read-artifact` logged `"WARN: artifact file not found"` for a condition that kills the implement phase (0ms duration, no output). Changed to `"ERROR: artifact file not found ... — MCP tool was likely not called by the LLM"` to make severity and root cause clear.

## Changes

| File | Change |
|------|--------|
| `workflow/runner.clj` | Reorder duration-ms extraction; `when-let` → `if-let` for terminal event constructors |
| `phase/implement.clj` | Override agent metrics `:duration-ms` with wall-clock time |
| `phase/verify.clj` | Same duration-ms override |
| `phase/review.clj` | Same duration-ms override |
| `phase/plan.clj` | Same duration-ms override |
| `agent/artifact_session.clj` | WARN → ERROR with root-cause hint |
| `workflow/run7_regression_test.clj` | 4 tests: extraction order, fallback event, duration override, stale phase cleanup |
| `agent/artifact_session_error_test.clj` | 2 tests: ERROR message content, nil return |

## Test plan

- [x] All 399 tests pass (6 new regression tests, 14 new assertions)
- [x] `duration-ms-extraction-order-test` — verifies wall-clock time wins over agent fallback 0
- [x] `phase-leave-overrides-agent-duration-test` — verifies leave functions write real duration
- [x] `publish-workflow-completed-fallback-test` — verifies event publishing doesn't silently fail
- [x] `phase-map-cleared-between-phases-test` — verifies stale `:redirect-to` doesn't leak (PR #268 regression guard)
- [x] `read-artifact-missing-file-logs-error-test` — verifies ERROR (not WARN) with MCP hint
- [ ] Run 8 of `finish-yc-mvp.spec.edn` to verify non-zero metrics in live output

### DIFF
--- a/components/workflow/src/ai/miniforge/workflow/runner.clj
+++ b/components/workflow/src/ai/miniforge/workflow/runner.clj
@@ -65,20 +65,14 @@
   [event-stream context]
   (when event-stream
     (try
-      (when-let [constructor (requiring-resolve 'ai.miniforge.event-stream.interface/workflow-started)]
+      (let [workflow-started (requiring-resolve 'ai.miniforge.event-stream.interface/workflow-started)]
         (publish-event! event-stream
-                        (constructor event-stream
-                                     (:execution/id context)
-                                     (select-keys (:execution/workflow context)
-                                                  [:workflow/id :workflow/version]))))
-      (catch Exception _
-        ;; Fallback to ad-hoc event if constructor not available
-        (publish-event! event-stream
-                        {:event/type :workflow/started
-                         :workflow/id (:execution/id context)
-                         :workflow/spec (select-keys (:execution/workflow context)
-                                                     [:workflow/id :workflow/version])
-                         :event/timestamp (java.time.Instant/now)})))))
+                        (workflow-started event-stream
+                                         (:execution/id context)
+                                         (select-keys (:execution/workflow context)
+                                                      [:workflow/id :workflow/version]))))
+      (catch Exception e
+        (println "Warning: Failed to publish workflow-started event:" (ex-message e))))))
 
 (defn publish-workflow-completed!
   "Publish workflow completed/failed event using N3-compliant constructor."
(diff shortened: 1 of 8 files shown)


components/workflow/src/ai/miniforge/workflow/runner.clj
```diff
@@ -89,14 +89,25 @@
...
+          (if-let [constructor (requiring-resolve 'ai.miniforge.event-stream.interface/workflow-failed)]
             (publish-event! event-stream
                             (constructor event-stream wf-id
                                          {:message "Workflow failed"
-                                          :errors (:execution/errors context)})))
-          (when-let [constructor (requiring-resolve 'ai.miniforge.event-stream.interface/workflow-completed)]
+                                          :errors (:execution/errors context)}))
             (publish-event! event-stream
-                            (constructor event-stream wf-id status duration-ms)))))
+                            {:event/type :workflow/failed
```
H31MDALLR: Do we not have common event creation fns?

PR #281, opened 2026-03-08
Fix TUI live workflow refresh and detail view data

H31MDALLR:
## Summary

- **File subscription poll loop** silently died on first exception — moved error handling inside the loop so failures are per-iteration, not fatal. New workflows now appear within 2s without restarting the TUI.
- **Workflow detail view** showed empty data because it only read from the in-memory snapshot (which could be sparse if live events were lost). Added `reload-workflow-detail` side-effect that reads the full event file from disk when entering detail view.
- **Size column color thresholds** extracted from hardcoded values to `governance/risk.edn` config with 5 new tests.

## Test plan

- [ ] Start TUI, then run `bb miniforge run work/<spec>` in another terminal — workflow should appear in TUI within 2 seconds
- [ ] Enter workflow detail while it's running — phases, agent output, and evidence should populate
- [ ] Verify Size column colors: green (≤200 LOC), neutral (200-500), yellow (>500), red (>1000)
- [ ] All 350 tests pass (1441 assertions, 0 failures)

### DIFF
--- a/components/config/resources/config/governance/digests.edn
+++ b/components/config/resources/config/governance/digests.edn
@@ -1,4 +1,4 @@
 {:readiness            "sha256:29dd89ac0172c8def994aad6498ed5ad04ffa94e6362df9d9a019ed9025b8027"
- :risk                 "sha256:e4950cd432f75430f2b42404497d66026605762bd1d20b3981715463f4973e55"
+ :risk                 "sha256:98b1724f02d3a26bde61321c5a57dba21930f9f2d7225a7132d897702c524217"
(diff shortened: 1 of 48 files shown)


H31MDALLR:
Review summary for repo rules:

I do not have new code-level findings on this follow-up. The main issues I called out on #280 look addressed here, and the new regression coverage is materially better.

I ran `bb test components/tui-views components/config` against the PR head and got: `381 tests`, `1540 assertions`, `0 failures`, `0 errors`.

The remaining repo-rule issue I still see is `721-pr-documentation`: I do not see the required PR doc for branch `fix/tui-live-workflow-refresh`. This PR was opened on March 8, 2026 in America/Los_Angeles time, so I would expect `docs/pull-requests/2026-03-08-fix-tui-live-workflow-refresh.md` in the diff.

PR #307, opened 2026-03-13
refactor: split CLI app identity from shared base

H31MDALLR:
# CLI App Profile Split

## Layer

Infrastructure

## Depends on

- #306 (`codex/kernel-project-split`) — merged

## What this adds

- moves CLI app identity and filesystem layout into resource-driven profiles
- keeps `miniforge` as the default/base profile
- gives `workflow-kernel` its own CLI name, home directory, and help examples
- makes project-owned CLI profile resources override base defaults explicitly instead of relying on classpath
  enumeration order

## Strata affected

- `bases/cli` — new app profile loader and path helpers
- `projects/workflow-kernel` — kernel-specific CLI identity override
- CLI tests and kernel integration tests — coverage for default and override behavior

## Why

- the shared CLI base was still hardcoding flagship branding and `~/.miniforge` paths even after the runtime and project
  seams were extracted
- the kernel project needs a separate app identity so the governed workflow engine can be renamed independently of the
  flagship software-factory product
- project-level CLI identity should be configuration data owned by the app/project layer, not literals embedded in the
  shared base

## Verification

- `clojure -M:dev:test -e "(require 'ai.miniforge.cli.app-config-test 'ai.miniforge.cli.main-test
  'ai.miniforge.cli.workflow-runner.display-test) ..."`
- `clojure -M -e "(require 'ai.miniforge.workflow.kernel-loader-integration-test) ..."` from `projects/workflow-kernel`
- `bb test`
- `bb test:integration`
- `bb build:kernel`
- `bb build:cli`
- `bb build:tui`
- `bb pre-commit`


bases/cli/src/ai/miniforge/cli/app_config.clj
```diff
@@ -0,0 +1,116 @@
...
+
+(defn command-string
+  "Build a CLI command string prefixed with the active binary name."
+  [& parts]
+  (str/join " " (cons (binary-name) (remove str/blank? parts))))
+
+;------------------------------------------------------------------------------ Layer 2
+;; Filesystem layout helpers
+
+(defn config-path []
```
H31MDALLR, line 93: These should be config. (EDN) and not hard coded.

bases/cli/src/ai/miniforge/cli/config.clj
```diff
@@ -263,13 +264,13 @@
     (if-not backend-str
       (do
         (print-error "Missing backend argument")
-        (println "Usage: miniforge config backend <name>")
+        (println "Usage:" (app-config/command-string "config backend <name>"))
```
H31MDALLR, line 267: All this English is going to come back to bite us if I ever have foreign customers. Consider whether we should set up for multilingual now with one language, or if it's okay to just wing it.

PR #317, opened 2026-03-15
feat: add repo-index component with scanner and repo map

H31MDALLR:
## Summary

- Add new `components/repo-index/` component with git-backed scanner (`git ls-tree`), token-budgeted repo map generator, and EDN storage layer
- Wire repo map into the implement phase so agents get a table of contents (~500 tokens) instead of blind file loading (~15k tokens)
- Update implementer agent prompt to describe and include the Repository Map

## Token Impact

| Before | After |
|--------|-------|
| 25 files × ~200 lines = ~15k tokens | ~500 token repo map + only files-in-scope |

Self-test on miniforge: 1,173 files indexed, 253k lines, 8 languages detected. 500-token budget → 30 files shown; 2000 tokens → 102 files.

## Note on PR size

This is 943 LOC (source + test) which exceeds the 400-line convention. It combines the 3 sub-PRs from the plan (scanner/schema scaffold, repo-map/storage, implement-phase wiring) because they form a single shippable unit — the repo map is useless without the scanner, and both are useless without the phase wiring.

## Test plan

- [x] 9 unit tests, 1,443 assertions (schema validation, scanner, repo map budget, file retrieval)
- [x] All 24 changed-brick test namespaces pass individually (206 tests, 2,438 assertions)
- [x] `implement.clj` and `implementer.clj` compile and load cleanly
- [ ] Dogfood: run a work spec and compare token usage before/after

See `docs/pull-requests/2026-03-14-feat-repo-index-scanner-repomap.md` for full details.


components/repo-index/src/ai/miniforge/repo_index/interface.clj
```diff
@@ -0,0 +1,190 @@
...
+       (try
+         (let [f (io/file (:repo-root index) file-path)]
+           (when (.isFile f)
+             (let [all-lines (str/split-lines (slurp f))
+                   line-count (count all-lines)
+                   truncated? (> line-count max-lines)
+                   kept-lines (if truncated?
+                                (take max-lines all-lines)
+                                all-lines)]
+               {:path file-path
```
H31MDALLR: If this is the *only* place we will create a file structure then this is OK. The moment we get a second, extract to a factory fn.

components/repo-index/src/ai/miniforge/repo_index/repo_map.clj
```diff
@@ -0,0 +1,172 @@
...
+                         {:path path
+                          :lang language
+                          :lines lines
+                          :size size})
+                       sorted)
+         ;; Group by top directory
+         grouped (group-by #(top-directory (:path %)) entries)
+         dir-order (sort (keys grouped))
+         ;; Build text incrementally, respecting token budget
+         header (str "## Repository Map\n"
```
H31MDALLR: Strings belong in the loc edn's.

components/repo-index/src/ai/miniforge/repo_index/repo_map.clj
```diff
@@ -0,0 +1,172 @@
...
+                     "Files: " (count entries) " | "
+                     "Lines: " (reduce + 0 (map :lines entries)) " | "
+                     "Languages: " (str/join ", " (map (fn [[k v]] (str k " (" v ")"))
+                                                       (sort-by (comp - val) (:languages index))))
+                     "\n")
+         header-tokens (estimate-tokens header)
+         remaining-budget (- budget header-tokens)
+         ;; Walk directory groups, accumulating text until budget is reached
+         result (reduce
+                  (fn [{:keys [text tokens shown truncated?] :as acc} dir]
```
H31MDALLR: extract fn

components/repo-index/src/ai/miniforge/repo_index/repo_map.clj
```diff
@@ -0,0 +1,172 @@
...
+  ([index] (generate index {}))
+  ([index opts]
+   (let [budget (or (:token-budget opts) default-token-budget)
+         exclude-gen? (get opts :exclude-generated? true)
+         files (cond->> (:files index)
+                 exclude-gen? (remove :generated?))
+         ;; Sort by directory grouping, then path
+         sorted (sort-by :path files)
+         ;; Build entries
+         entries (mapv (fn [{:keys [path language lines size]}]
```
H31MDALLR: This looks like a file structure map like the one we were creating above in a different file. That means that we have two of these now and need to extract it to a factory function. Also extract this function to a function.

components/repo-index/src/ai/miniforge/repo_index/repo_map.clj
```diff
@@ -0,0 +1,172 @@
...
+                                                       (sort-by (comp - val) (:languages index))))
+                     "\n")
+         header-tokens (estimate-tokens header)
+         remaining-budget (- budget header-tokens)
+         ;; Walk directory groups, accumulating text until budget is reached
+         result (reduce
+                  (fn [{:keys [text tokens shown truncated?] :as acc} dir]
+                    (if truncated?
+                      acc
+                      (let [group-entries (get grouped dir)
```
H31MDALLR: When you do extract this function, decompose it as well. Even as an extracted function, it's too large. Should read like a pipeline.

components/repo-index/src/ai/miniforge/repo_index/repo_map.clj
```diff
@@ -0,0 +1,172 @@
...
+                          ;; Fits in budget
+                          {:text (str text "\n" group-text)
+                           :tokens new-tokens
+                           :shown (+ shown (count group-entries))
+                           :truncated? false}))))
+                  {:text "" :tokens 0 :shown 0 :truncated? false}
+                  dir-order)
+         full-text (str header (:text result))
+         total-tokens (+ header-tokens (:tokens result))]
+     {:tree-sha (:tree-sha index)
```
H31MDALLR: Create a factory function for this.

components/repo-index/src/ai/miniforge/repo_index/repo_map.clj
```diff
@@ -0,0 +1,172 @@
...
+                                                 {:rows [] :row-tokens 0}
+                                                 group-entries)]
+                            (if (seq (:rows partial-entries))
+                              {:text (str text "\n" group-header "\n" (str/join "\n" (:rows partial-entries)))
+                               :tokens (+ tokens (estimate-tokens group-header) (:row-tokens partial-entries))
+                               :shown (+ shown (count (:rows partial-entries)))
+                               :truncated? true}
+                              (assoc acc :truncated? true)))
+                          ;; Fits in budget
+                          {:text (str text "\n" group-text)
```
H31MDALLR: Should also be in a factory function.

components/agent/src/ai/miniforge/agent/implementer.clj
```diff
@@ -36,143 +37,194 @@
...
-
 (def implementer-system-prompt
   "System prompt for the implementer agent.
    Loaded from EDN resource for configurability."
   (delay (prompts/load-prompt :implementer)))
 
+;------------------------------------------------------------------------------ Layer 0
+;; Extension → language mapping
+
+(def ^:private ext->language
```
H31MDALLR: Wouldn't this be configuration? In an EDN file I don't know. I think language names are actually not localized because they're proper names and branded, so we might not need those in the localization EDN, but it's definitely a configuration.

components/agent/src/ai/miniforge/agent/implementer.clj
```diff
@@ -36,143 +37,194 @@
...
-          {:valid? true :errors nil})))))
+  (if-not (m/validate CodeArtifact artifact)
+    {:valid? false
+     :errors (schema/explain CodeArtifact artifact)}
+    (let [files (:code/files artifact)
+          empty-creates (find-empty-creates files)
+          duplicate-paths (find-duplicate-paths files)]
+      (cond
+        (seq empty-creates)
+        {:valid? false
```
H31MDALLR: It seems like it should be a factory function that generates errors. If this can't use one of our existing anomaly patterns, it would be a new one.

components/agent/src/ai/miniforge/agent/implementer.clj
```diff
@@ -184,73 +236,166 @@
...
-;; make-fallback-code removed — silent fallback masks real failures,
-;; prevents retry/repair from working, and short-circuits checkpoint resume.
+;------------------------------------------------------------------------------ Layer 1
+;; Artifact repair
+
+(defn- ensure-required-fields
+  "Ensure a file entry has path, content, and action."
+  [f]
+  (cond-> f
+    (not (:path f))    (assoc :path "src/generated/unknown.clj")
```
H31MDALLR: Is this better than null as a path?

components/repo-index/src/ai/miniforge/repo_index/interface.clj
```diff
@@ -0,0 +1,192 @@
...
+   Returns {:path :content :lines :truncated?} or nil on error."
+  [repo-root file-path max-lines]
+  (try
+    (let [f (io/file repo-root file-path)]
+      (when (.isFile f)
+        (let [all-lines (str/split-lines (slurp f))
+              line-count (count all-lines)
+              truncated? (> line-count max-lines)
+              kept-lines (if truncated? (take max-lines all-lines) all-lines)]
+          {:path file-path
```
H31MDALLR: This feels like it should be part of the factory pattern that creates this structure because I've seen it before.

components/repo-index/src/ai/miniforge/repo_index/scanner.clj
```diff
@@ -0,0 +1,237 @@
...
+  "Walk git tree via `git ls-tree` to build a repo index.
+
+   Layer 1 — depends on schema (Layer 0)."
+  (:require [clojure.java.io :as io]
+            [clojure.string :as str]))
+
+;------------------------------------------------------------------------------ Layer 0
+;; Language detection (extension-based)
+
+(def ^:private extension->language
```
H31MDALLR: It feels very much like config, which we could pull out to an env file. And update without code updates.

H31MDALLR: Fixed. Moved to `resources/config/repo-index/scanner.edn` — loaded at startup via `load-scanner-config`. Extension mappings, special filenames, and generated-path patterns are all config now.

components/repo-index/src/ai/miniforge/repo_index/repo_map.clj
H31MDALLR: Fixed. Table header/separator and all repo-map labels moved to messages/en.edn. format-group-header calls messages/t.

components/repo-index/src/ai/miniforge/repo_index/repo_map.clj
H31MDALLR: Fixed. Extracted to render-directory-group, which delegates to render-group-rows, render-group-text, fit-partial-group, and append-full-group.

components/repo-index/src/ai/miniforge/repo_index/repo_map.clj
H31MDALLR: Fixed. Created factory/->repo-map-entry. Both this and the FileRecord map go through factory functions in factory.clj.

components/repo-index/src/ai/miniforge/repo_index/repo_map.clj
H31MDALLR: Fixed. render-directory-group reads as a pipeline: compute group metadata → render-group-text → budget check → fit-partial-group or append-full-group. Inner row-fitting extracted to render-group-rows.

components/repo-index/src/ai/miniforge/repo_index/repo_map.clj
H31MDALLR: Fixed. Created factory/->repo-map-slice. generate now calls it instead of constructing the map inline.

components/repo-index/src/ai/miniforge/repo_index/repo_map.clj
H31MDALLR: Fixed. Created factory/->render-acc and factory/render-acc-with. All accumulator construction goes through these.

components/agent/src/ai/miniforge/agent/implementer.clj
H31MDALLR: Fixed. Moved to resources/config/repo-index/implementer.edn, loaded via load-ext->language at startup. Agreed these aren't localized — they're config.

components/agent/src/ai/miniforge/agent/implementer.clj
H31MDALLR: Fixed. validate-code-artifact now uses schema/valid and schema/invalid from the existing anomaly patterns in schema.interface.

components/agent/src/ai/miniforge/agent/implementer.clj
H31MDALLR: Fixed. ensure-required-fields now returns nil for entries missing a path (dropping them). The repair pipeline filters out nils.

components/repo-index/src/ai/miniforge/repo_index/interface.clj
H31MDALLR: Fixed. Created factory/->file-content. read-file-limited uses it — single source of truth for this map shape.

components/repo-index/src/ai/miniforge/repo_index/interface.clj
H31MDALLR: We now have two call sites, so extracted to factory/->file-content.

PR #323, opened 2026-03-16
fix: increase max-turns to 10, raise context budgets to match LLM capacity

H31MDALLR:
## Summary

- Increase `--max-turns` from 5 to 10 (agents need 8-12 turns for complex tasks)
- Raise context-pack budgets from 1.2K-3.5K to 20K-100K tokens per phase
- Agents were exploring because they only got 1-2 files in a 3.5K budget; now they get 10-20+ files, reducing need to explore

## Why the old budgets were wrong

| Phase | Old | New | Rationale |
|-------|-----|-----|-----------|
| plan | 1,200 | 20,000 | Planner needs to see repo structure + key files |
| implement | 3,500 | 100,000 | 3.5K fits ~2 files; complex tasks touch 10+ files |
| test | 2,200 | 60,000 | Tester needs full code artifact + existing tests |
| review | 2,800 | 40,000 | Reviewer needs code + test output + context |

Input tokens are cheap ($3-15/M). Each exploration turn the agent spends reading files costs $0.05-0.10 + 1-2 min. Pre-loading context eliminates those turns.

## Codex MCP status

Codex auto-discovers MCP from `.codex/config.toml` (which `artifact_session.clj` already writes). No CLI flags needed. Codex `exec` mode doesn't have built-in file tools like Claude CLI, so it should use MCP tools more reliably.

## Test plan

- [x] Context-pack, implementer, tester, LLM tests all pass
- [ ] Re-run YC MVP spec to verify complex tasks complete

### DIFF
--- a/components/agent/src/ai/miniforge/agent/implementer.clj
+++ b/components/agent/src/ai/miniforge/agent/implementer.clj
@@ -383,7 +383,7 @@
                           :mcp-allowed-tools (:mcp-allowed-tools session)
                           :supervision (:supervision session)
                           :budget-usd budget-usd
-                          :max-turns 5}]
+                          :max-turns 10}]
             (if on-chunk
               (llm/chat-stream llm-client user-prompt on-chunk
                                (merge {:system effective-system-prompt} mcp-opts))
--- a/components/agent/src/ai/miniforge/agent/implementer.clj
+++ b/components/agent/src/ai/miniforge/agent/implementer.clj
@@ -43,6 +43,10 @@
    Loaded from EDN resource for configurability."
   (delay (prompts/load-prompt :implementer)))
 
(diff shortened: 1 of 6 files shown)


components/agent/src/ai/miniforge/agent/implementer.clj
H31MDALLR, line 387: Turns should be in config so we do not have to ship code to update these.

H31MDALLR: Fixed. Both implementer and tester now read `:prompt/max-turns` from their EDN prompt configs (`resources/prompts/implementer.edn` and `resources/prompts/tester.edn`) instead of hardcoding the value. Falls back to 10 if the key is missing.

PR #330, opened 2026-03-25
feat: add human meta-loop decision checkpoints

H31MDALLR:
# feat: human meta-loop checkpoint model

## Overview

Introduces the first implementation slice for the human meta-loop spec:
a canonical normalized decision model that both the control-plane queue and
inner-loop escalation path can emit.

This does not build the inbox, rule miner, or delegated-mode runtime yet.
It establishes the shared artifact those later features need: a versioned,
structured checkpoint and episode shape.

The branch also includes a small reader-fix in `components/algorithms` for
`graph.clj`. That parse error was already present on `origin/main` after the
rebase and blocked `bb pre-commit` from completing, so it is included here to
keep the branch reviewable and commit hooks green.

## Motivation

The current repo has two adjacent but separate concepts:

- control-plane decisions for external or supervised agents
- loop escalation for retry-budget exhaustion

Both represent judgment checkpoints, but they use incompatible shapes.
That makes it hard to build one inbox, one audit trail, and one eventual
delegation-acquisition pipeline.

This PR starts by normalizing the data before expanding UI or automation.

### 1. New `decision` component

Adds a pure component for canonical decision artifacts:

- `DecisionCheckpoint`
- `DecisionEpisode`
- constructors and update helpers
- producer helpers for control-plane decisions and loop escalations

### 2. Control-plane queue enrichment

`components/control-plane` now keeps its existing `:decision/*` API while
attaching:

- `:decision/checkpoint`
- `:decision/episode`

That preserves current consumers while giving downstream features a shared
normalized representation.

### 3. Loop escalation bridge

`components/loop` now emits the same normalized checkpoint/episode model when
an inner loop escalates or aborts.

This makes loop escalation a first-class checkpoint producer instead of an
isolated prompt UX.

### 4. Reader fix for `algorithms/graph.clj`

Repairs a malformed DFS implementation in `components/algorithms` that broke
namespace loading under both Clojure and Babashka/GraalVM.

This is not part of the product feature itself, but it was required to satisfy
normal pre-commit discipline on the rebased branch.

## Testing Plan

Targeted namespace validation:

```bash
CLJ_CONFIG=/tmp/codex-clojure-config clojure -M:test -e "(require '[clojure.test :as t] '[ai.miniforge.decision.interface-test]) (let [r (t/run-tests 'ai.miniforge.decision.interface-test)] (System/exit (if (zero? (+ (:fail r) (:error r))) 0 1)))"

CLJ_CONFIG=/tmp/codex-clojure-config clojure -M:test -e "(require '[clojure.test :as t] '[ai.miniforge.loop.escalation-test]) (let [r (t/run-tests 'ai.miniforge.loop.escalation-test)] (System/exit (if (zero? (+ (:fail r) (:error r))) 0 1)))"

CLJ_CONFIG=/tmp/codex-clojure-config clojure -M:test -e "(require '[clojure.test :as t] '[ai.miniforge.control-plane.interface-test]) (let [r (t/run-tests 'ai.miniforge.control-plane.interface-test)] (System/exit (if (zero? (+ (:fail r) (:error r))) 0 1)))"
```

Full changed-bricks and compatibility validation:

```bash
CLJ_CONFIG=/tmp/codex-clojure-config bb pre-commit
```

## Deployment Plan

Safe incremental change:

- legacy `:decision/*` fields remain intact
- new checkpoint/episode data is additive
- no UI migration is required yet

## Related Issues/PRs

- Human meta-loop spec provided for this branch: `miniforge-human-meta-loop-spec.md`
- Existing escalation UX foundation:
  `docs/pull-requests/2026-01-24-feat-human-escalation-ux.md`

## Checklist

- [x] Canonical checkpoint schema added
- [x] Canonical episode schema added
- [x] Control-plane decisions emit normalized data
- [x] Loop escalation emits normalized data
- [x] Targeted tests passing
- [ ] Unified inbox UI
- [ ] Episode persistence/warehouse
- [ ] Similar-decision retrieval
- [ ] Delegation rule mining

### DIFF
--- a/components/algorithms/src/ai/miniforge/algorithms/graph.clj
+++ b/components/algorithms/src/ai/miniforge/algorithms/graph.clj
@@ -28,6 +28,52 @@
 ;------------------------------------------------------------------------------ Layer 0
 ;; Core DFS traversal
 
+(defn- dfs-traverse
+  [graph node-id path visited visiting get-deps-fn on-visit-fn on-cycle-fn on-missing-fn]
+  (cond
+    ;; Already visited - continue
+    (contains? visited node-id)
+    [visited nil]
+
+    ;; Currently visiting - cycle detected
+    (contains? visiting node-id)
+    (let [result (on-cycle-fn node-id (conj path node-id) visited visiting)]
+      [visited result])
+
+    ;; Visit this node
+    :else
+    (let [node (get graph node-id)]
+      (if-not node
+        ;; Node not in graph
+        (let [result (on-missing-fn node-id visited visiting)]
+          [visited result])
+        ;; Process node
+        (let [result (on-visit-fn node-id node path visited visiting)]
+          (if result
+            ;; Halt with result
+            [visited result]
+            ;; Continue with dependencies
+            (let [deps (get-deps-fn node)
+                  visiting' (conj visiting node-id)]
+              (loop [remaining-deps deps
+                     current-visited visited
+                     current-result nil]
+                (if (or current-result (empty? remaining-deps))
+                  [(conj current-visited node-id) current-result]
+                  (let [[next-visited next-result]
+                        (dfs-traverse graph
+                                      (first remaining-deps)
+                                      (conj path node-id)
+                                      current-visited
+                                      visiting'
+                                      get-deps-fn
+                                      on-visit-fn
+                                      on-cycle-fn
+                                      on-missing-fn)]
+                    (recur (rest remaining-deps)
+                           next-visited
+                           next-result)))))))))))
+
 (defn dfs
   "Simple canonical depth-first search traversal.
 
@@ -62,51 +108,22 @@
          (fn [id visited visiting] nil))           ; on-missing"
   [graph start-ids get-deps-fn on-visit-fn on-cycle-fn on-missing-fn]
   (let [start-ids (if (coll? start-ids) start-ids [start-ids])]
-    (letfn [(traverse [node-id path visited visiting]
-              (cond
-                ;; Already visited - continue
-                (contains? visited node-id)
-                [visited nil]
-
-                ;; Currently visiting - cycle detected
-                (contains? visiting node-id)
-                (let [result (on-cycle-fn node-id (conj path node-id) visited visiting)]
-                  [visited result])
-
-                ;; Visit this node
-                :else
-                (let [node (get graph node-id)]
-                  (if-not node
-                    ;; Node not in graph
-                    (let [result (on-missing-fn node-id visited visiting)]
-                      [visited result])
-                    ;; Process node
-                    (let [result (on-visit-fn node-id node path visited visiting)]
-                      (if result
-                        ;; Halt with result
-                        [visited result]
-                        ;; Continue with dependencies
-                        (let [deps (get-deps-fn node)
-                              visiting' (conj visiting node-id)]
-                          (loop [remaining-deps deps
-                                 v visited
-                                 result nil]
-                            (if (or result (empty? remaining-deps))
-                              [(conj v node-id) result]
-                              (let [[v' result'] (traverse (first remaining-deps)
-                                                           (conj path node-id)
-                                                           v
-                                                           visiting')]
-                                (recur (rest remaining-deps) v' result'))))))))))]
-
-      ;; Process all start nodes
-      (loop [remaining-starts start-ids
-             visited #{}
-             result nil]
-        (if (or result (empty? remaining-starts))
-          [visited result]
-          (let [[visited' result'] (traverse (first remaining-starts) [] visited #{})]
-            (recur (rest remaining-starts) visited' result')))))))
+    (loop [remaining-starts start-ids
+           visited #{}
+           result nil]
+      (if (or result (empty? remaining-starts))
+        [visited result]
+        (let [[visited' result']
+              (dfs-traverse graph
+                            (first remaining-starts)
+                            []
+                            visited
+                            #{}
+                            get-deps-fn
+                            on-visit-fn
+                            on-cycle-fn
+                            on-missing-fn)]
+          (recur (rest remaining-starts) visited' result'))))))
 
 ;------------------------------------------------------------------------------ Layer 1
 ;; DFS-based helper functions
--- a/components/algorithms/src/ai/miniforge/algorithms/graph.clj
+++ b/components/algorithms/src/ai/miniforge/algorithms/graph.clj
@@ -28,6 +28,46 @@
 ;------------------------------------------------------------------------------ Layer 0
 ;; Core DFS traversal
 
+(declare dfs-traverse)
+
+(defn- dfs-traverse-deps
+  [graph node-id deps path visited visiting get-deps-fn on-visit-fn on-cycle-fn on-missing-fn]
+  (loop [remaining-deps deps
+         current-visited visited
+         current-result nil]
+    (if (or current-result (empty? remaining-deps))
+      [(conj current-visited node-id) current-result]
+      (let [[next-visited next-result]
+            (dfs-traverse graph
+                          (first remaining-deps)
+                          (conj path node-id)
+                          current-visited
+                          visiting
+                          get-deps-fn
+                          on-visit-fn
+                          on-cycle-fn
+                          on-missing-fn)]
+        (recur (rest remaining-deps)
+               next-visited
+               next-result)))))
+
+(defn- dfs-visit-node
+  [graph node-id path visited visiting get-deps-fn on-visit-fn on-cycle-fn on-missing-fn]
+  (if-let [node (get graph node-id)]
+    (if-let [result (on-visit-fn node-id node path visited visiting)]
+      [visited result]
+      (dfs-traverse-deps graph
+                         node-id
+                         (get-deps-fn node)
+                         path
+                         visited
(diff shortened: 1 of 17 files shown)


components/algorithms/src/ai/miniforge/algorithms/graph.clj
```diff
@@ -28,6 +28,52 @@
...
+          (if result
+            ;; Halt with result
+            [visited result]
+            ;; Continue with dependencies
+            (let [deps (get-deps-fn node)
+                  visiting' (conj visiting node-id)]
+              (loop [remaining-deps deps
+                     current-visited visited
+                     current-result nil]
+                (if (or current-result (empty? remaining-deps))
```
H31MDALLR: It seems unreasonably nested and obtuse for a DFS algorithm. Refactor using Clojure idioms, and let's see if we can get this to a much more maintainable DFS algorithm. It currently has at least five layers of nested ifs and a loop. Adhere to the simple, made easy and stratified design and also the other rule sets we have.

components/control-plane/src/ai/miniforge/control_plane/decision_queue.clj
```diff
@@ -109,13 +116,26 @@
...
+                                  :choice :choose-option
+                                  :input :approve-with-constraints
+                                  :confirmation :approve
+                                  :approve)
+                          :value resolution
+                          :authority-role :human}
+                   comment (assoc :rationale comment))
+        checkpoint (decision/resolve-checkpoint (:decision/checkpoint decision) response)
+        episode (decision/update-episode (:decision/episode decision) checkpoint)]
+    (assoc decision
```
H31MDALLR: I think this might be better as a function that returns this map. 

```clojure
(resolve-decision resolution comment checkpoint episode)
```

..etc.

components/control-plane/src/ai/miniforge/control_plane/orchestrator.clj
```diff
@@ -43,6 +45,47 @@
...
+    (or (:task status-update) (:agent/task status-update))
+    (assoc :task (or (:task status-update) (:agent/task status-update)))
+
+    (or (:metrics status-update) (:agent/metrics status-update))
+    (assoc :metrics (or (:metrics status-update) (:agent/metrics status-update)))))
+
+(defn- normalize-decision-opts
+  [opts]
+  (cond-> {}
+    (or (:type opts) (:decision/type opts))
```
H31MDALLR: Why do we have an or here? This feels like fallback behavior. We have no users. Do the right thing without adding complexity to maintain behavior no one will use.

components/control-plane/src/ai/miniforge/control_plane/orchestrator.clj
```diff
@@ -43,6 +45,47 @@
...
+    (assoc :status (or (:status status-update) (:agent/status status-update)))
+
+    (or (:task status-update) (:agent/task status-update))
+    (assoc :task (or (:task status-update) (:agent/task status-update)))
+
+    (or (:metrics status-update) (:agent/metrics status-update))
+    (assoc :metrics (or (:metrics status-update) (:agent/metrics status-update)))))
+
+(defn- normalize-decision-opts
+  [opts]
```
H31MDALLR: Use destructuring and move the whichever one of these accesses survives up into the signature.

components/decision/src/ai/miniforge/decision/core.clj
```diff
@@ -0,0 +1,182 @@
...
+                          :decision-class (get decision-type->class decision-type :implementation-pattern-choice)
+                          :summary summary}
+                   (seq options) (assoc :alternatives
+                                        (mapv option->alternative (range) options)))
+        uncertainty (cond-> {:class :agent-request
+                             :reason "External or delegated agent requested human judgment."}
+                      (some? (:agent-confidence opts))
+                      (assoc :agent-confidence (:agent-confidence opts)))]
+    (create-checkpoint
+     {:source {:kind :control-plane-agent
```
H31MDALLR: Why are we creating hand-built maps to pass to a function that is a factory function for creating a checkpoint? Is this a common context map? If so, we should have a function that takes in values, returns the map, and then the create checkpoint can consume that.

components/decision/src/ai/miniforge/decision/core.clj
```diff
@@ -0,0 +1,182 @@
...
+        task-info (cond-> {:kind :loop-escalation
+                           :goal summary}
+                    (:task/id task) (assoc :task-id (:task/id task)))
+        proposal (cond-> {:action-type :provide-guidance
+                          :decision-class :repair-escalation
+                          :summary summary}
+                   (:artifact/path artifact) (assoc :files [(:artifact/path artifact)])
+                   diff-summary (assoc :diff-summary diff-summary))]
+    (create-checkpoint
+     {:source {:kind :loop-escalation
```
H31MDALLR: I'm here. Find this pattern and let's make it simpler so we're not constantly building hand-built maps that we then pass to a factory function. Instead, we do this without the extra hop, or we do it with a factory function that generates the context this function is looking for.

components/decision/src/ai/miniforge/decision/core.clj
```diff
@@ -0,0 +1,182 @@
...
+               :loop-id (:loop/id loop-state)}
```
H31MDALLR: Also make sure this context has a mail schema, and we validate it. If it's internal, we don't have to validate it inside this component, but if it comes from an interface boundary that needs validation

components/loop/src/ai/miniforge/loop/escalation.clj
```diff
@@ -142,15 +161,30 @@
...
       ;; No escalation function, default to abort
-      {:escalated true
-       :action :abort
-       :reason "No escalation function provided"})))
+      (let [result {:action :abort
+                    :reason "No escalation function provided"}
+            resolved-checkpoint (decision/resolve-checkpoint checkpoint
+                                                             (result->response result))
+            updated-episode (decision/update-episode episode resolved-checkpoint)]
+        {:escalated true
```
H31MDALLR: Also should probably be a factory function:

```clojure
(abort :not-found resolved-checkpoint updated-episode)
```

H31MDALLR: Done in commit 1773e78. The escalation path now uses small result factories, `escalated-result` and `abort-escalation`, instead of repeating raw output assembly inline.

components/decision/src/ai/miniforge/decision/core.clj
H31MDALLR: Refactored this in commit 1773e78. Control-plane checkpoint creation now goes through narrower helpers like `control-plane-proposal`, `control-plane-uncertainty`, `control-plane-context`, and `control-plane-checkpoint-input` before the final constructor, instead of assembling one large ad hoc map inline.

components/algorithms/src/ai/miniforge/algorithms/graph.clj
H31MDALLR: Refactored the DFS repair into smaller helpers in commit 1773e78. The traversal is now split across `dfs-traverse-deps`, `dfs-visit-node`, and `dfs-traverse`, which removes the deeply nested control flow while keeping the behavior and tests intact.

components/control-plane/src/ai/miniforge/control_plane/orchestrator.clj
H31MDALLR: Removed the fallback normalization in commit 1773e78. The orchestrator now accepts the documented control-plane adapter shape directly: `:status`, `:task`, and `:metrics` for poll updates, and the tests were updated to match that single contract.

components/control-plane/src/ai/miniforge/control_plane/orchestrator.clj
H31MDALLR: Applied that simplification in commit 1773e78. The surviving keys are destructured at the call site in `run-poll-pass`, and I removed the extra normalization helper entirely.

components/decision/src/ai/miniforge/decision/core.clj
H31MDALLR: Added explicit context schemas and interface-boundary validation in commit 1773e78. `decision.spec` now defines `ControlPlaneContext`, `LoopEscalationContext`, and `DecisionContext`, and the interface validates those through the checkpoint constructors rather than leaving context as an unstructured map.

components/decision/src/ai/miniforge/decision/core.clj
H31MDALLR: Applied across both checkpoint producers in commit 1773e78. Loop escalation now mirrors the same pattern with `loop-escalation-summary`, `loop-escalation-task`, `loop-escalation-proposal`, `loop-escalation-context`, and `loop-escalation-checkpoint-input` so the generic checkpoint constructor is only consuming shaped input, not giant inline maps.

components/control-plane/src/ai/miniforge/control_plane/decision_queue.clj
H31MDALLR: Pulled the final decision assoc into a dedicated helper in commit 1773e78. `resolve-decision` now builds the response and checkpoint state, then delegates the output map assembly to `resolved-decision-record`.

PR #338, opened 2026-03-27
Modify inject-knowledge in store.clj to return a manifest alongside...

H31MDALLR:

### DIFF
--- a/components/knowledge/src/ai/miniforge/knowledge/store.clj
+++ b/components/knowledge/src/ai/miniforge/knowledge/store.clj
@@ -343,29 +343,36 @@
   (get default-agent-manifests role
        {:agent-role role :types [:rule] :max-zettels 10}))
 
+(def ^:private manifest-score-weights
+  "Weights for computing manifest entry relevance scores.
+   Each weight controls how much a particular match type contributes
+   to the total score used for ranking and observability."
+  {:tag-match   1.0   ; per matching tag
+   :dewey-match 1.0   ; dewey classification prefix match
+   :type-match  0.5}) ; zettel type match
+
 (defn- compute-manifest-entry
   "Compute a manifest entry for a zettel matched during knowledge injection.
 
-   Scores each zettel based on how well it matches the query criteria:
-   - +1.0 per matching tag
-   - +1.0 for dewey prefix match
-   - +0.5 for type match
+   Scores each zettel based on how well it matches the query criteria
+   using weights from `manifest-score-weights`.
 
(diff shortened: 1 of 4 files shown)


components/knowledge/src/ai/miniforge/knowledge/store.clj
```diff
@@ -343,6 +343,72 @@
...
+(defn- compute-manifest-entry
+  "Compute a manifest entry for a zettel matched during knowledge injection.
+
+   Scores each zettel based on how well it matches the query criteria:
+   - +1.0 per matching tag
+   - +1.0 for dewey prefix match
+   - +0.5 for type match
+
+   Returns map with :id, :title, :role, :tags-matched, and :score."
+  [zettel agent-role query-tags dewey-prefixes query-types]
```
H31MDALLR, line 361: Remove the magic numbers; use good names for them, and move those to a configuration. We should not have just raw numbers sitting in here.

components/knowledge/src/ai/miniforge/knowledge/store.clj
```diff
@@ -343,6 +343,72 @@
...
+   Returns:
+   {:zettels  [zettel...]
+    :manifest [{:id uuid, :title string, :role keyword,
+                :tags-matched [keyword...], :score number}]}"
+  [store agent-role & [context]]
+  (let [agent-manifest  (get-agent-manifest agent-role)
+        base-query      {:dewey-prefixes (:dewey-prefixes agent-manifest)
+                         :include-types  (:types agent-manifest)
+                         :limit          (:max-zettels agent-manifest)}
+        context-tags    (or (:tags context) [])
```
H31MDALLR: Don't use the (or (:k m) default) pattern. Use (get m k default).

components/knowledge/src/ai/miniforge/knowledge/store.clj
H31MDALLR: Fixed in f695498 — extracted magic numbers into `manifest-score-weights` config map with named keys (`:tag-match`, `:dewey-match`, `:type-match`).

components/knowledge/src/ai/miniforge/knowledge/store.clj
H31MDALLR: Fixed in f695498 — replaced all `(or (:k m) default)` with `(get m k default)` throughout `inject-knowledge-with-manifest` and `compute-manifest-entry`.

PR #388, opened 2026-04-03
Define open Malli schemas for the 6 v1 supervisory entities in a new

H31MDALLR:


H31MDALLR:
Superseded by #390 which contains additional commits and the complete implementation.

PR #413, opened 2026-04-06
refactor: proper Polylith distribution — bundle mcp-context-server and lsp-mcp-bridge into mf

H31MDALLR:
## Summary

- **mcp-context-server** promoted from component → base; standalone project removed; bundled into `miniforge-core` and `mf` binary
- **lsp-mcp-bridge** given same treatment: bundled into `miniforge-core`, MCP config generation updated from `bb lsp-mcp-bridge` → `mf lsp-mcp-bridge`
- **`mf context-server`** and **`mf lsp-mcp-bridge`** added as CLI subcommands (internal, spawned by agent/MCP clients)
- **`mf lsp status/install/setup`** added as user-facing CLI subcommands
- **`server-command`** given a proper 4-step resolution chain: `MINIFORGE_MCP_CMD` env → bb task (dev) → `mf context-server` (prod) → `miniforge context-server` (fallback)

Distribution model is now correct: tools invisible to users (MCP servers) are bundled in the `mf` binary. The TUI remains standalone due to its JVM/GraalVM incompatibility.

## Test plan

- [ ] `mf context-server --artifact-dir /tmp/test` starts the JSON-RPC server
- [ ] `mf lsp-mcp-bridge` starts the LSP bridge (needs `MINIFORGE_PROJECT_DIR`)
- [ ] `mf lsp status` prints LSP server installation status
- [ ] `mf lsp install clojure` installs the Clojure LSP server
- [ ] `mf lsp setup --claude-code` writes `.mcp.json` with `mf lsp-mcp-bridge` command
- [ ] `mf lsp setup --claude-desktop` writes Claude Desktop config with `mf lsp-mcp-bridge`
- [ ] Existing `bb lsp-mcp-bridge` and `bb lsp:setup` tasks still work in dev mode


bases/cli/src/ai/miniforge/cli/main.clj
```diff
@@ -227,6 +227,32 @@
         start-server! (requiring-resolve 'ai.miniforge.mcp-context-server.interface/start-server)]
```
H31MDALLR: Requiring resolve is an anti-pattern. No need for this. Import the correct poly-lith components. Make sure we don't have any cycles. Requiring a resolve is something we've removed from our codebase. Don't add it back in.

bases/cli/src/ai/miniforge/cli/main.clj
```diff
@@ -227,6 +227,32 @@
...
+    (status!)))
+
+(defn lsp-install-cmd [m]
+  (let [install! (requiring-resolve 'ai.miniforge.lsp-mcp-bridge.tasks/install)]
+    (install! (:args m))))
+
+(defn lsp-setup-cmd [m]
+  (let [opts     (get-opts m)
+        args     (cond-> []
+                   (:claude-code opts)    (conj "--claude-code")
```
H31MDALLR: This really looks like we should be querying for configuration instead of having an explicit test for every single kind of agent that we want to add here. That way, the user could extend this without us having to ship new code.

components/agent/resources/prompts/tester.edn
H31MDALLR, line 121: How tests are written and where they land is actually policy. Not sure how to hand that tension with the need for a canonical tester persona.

bases/cli/src/ai/miniforge/cli/main.clj
H31MDALLR: Fixed in 06cc9b8. All `requiring-resolve` calls replaced with direct invocations via static requires in the `(:require ...)` block: `tool-supervisor`, `mcp-context-server`, `lsp-bridge`, and `lsp-tasks`. No cycles — these are leaf bases/components with no back-references to cli.

bases/cli/src/ai/miniforge/cli/main.clj
H31MDALLR: Fixed in 06cc9b8. Added `client-registry` map in `tasks.clj` keyed by CLI flag (`--claude-code`, `--claude-desktop`, `--codex`, `--codex-cli`, `--codex-desktop`). The `setup` fn now dispatches via registry lookup. New clients are added by extending the map — no code changes to the dispatch logic. The `lsp-setup-cmd` in `main.clj` now just passes `(:args m)` through and the per-client boolean spec is gone from the dispatch entry.

components/agent/resources/prompts/tester.edn
H31MDALLR: Agreed — test placement is genuinely a policy question (which persona owns coverage for a given component) rather than something the canonical tester persona should encode. No code change in this PR; worth a separate design discussion on whether tester.edn or a gate-level policy is the right place to express that contract.

PR #414, opened 2026-04-06
fix: prevent concurrent git worktree failures in DAG sub-workflow execution

H31MDALLR:
## Summary

- **Root cause**: When the DAG orchestrator runs 4 tasks concurrently via `futures`, each task calls `acquire-execution-environment!` → `WorktreeExecutor.acquire-environment!` → `git worktree add`. Git uses a `.git/config.lock` file internally; concurrent calls from multiple threads race on this lock and fail silently.
- **Broken fallback**: The old fallback on failure was `git worktree add /path main` (without `-b`), which **always fails** when `main` is already checked out in the primary worktree. This left all sub-workflows with a nil `:execution/worktree-path`.
- **Cascade**: Implement phase's `resolve-worktree-path` throws before any LLM call (0 tokens). `propagate-failures` cascades to all dependents. Result: all 28 tasks fail with `tasks-completed 0`.

## Changes

**`worktree.clj`** — two fixes:
1. Add `worktree-lock` (JVM `Object`) and wrap `git worktree add` in `locking`. Tasks still execute in parallel; only worktree _creation_ is serialized (fast, ~50ms each).
2. Fix the fallback chain: on first failure, clean up stale branch/directory from prior run and retry, then use `--detach` as a final last resort instead of the broken re-checkout of `main`.

**`runner.clj`** — add a `WARN` println in `build-env-record` when worktree acquisition returns an error result, so the failure is visible rather than silently producing `nil`.

## Test plan

- [x] All 521 tests pass (dag-executor, workflow, pr-lifecycle, spec-parser)
- [ ] Dogfood a 5+ task plan to verify DAG tasks complete with non-zero tokens

### DIFF
--- a/bb.edn
+++ b/bb.edn
@@ -443,7 +443,6 @@
                  "components/dag-executor/resources"
                  "components/task-executor/src"
                  "components/pr-lifecycle/src"
-                "components/pr-lifecycle/resources"
                  "components/release-executor/src"
                  "components/evidence-bundle/src"
                  "components/evidence-bundle/resources"
--- a/bb.edn
+++ b/bb.edn
@@ -443,6 +443,7 @@
                  "components/dag-executor/resources"
                  "components/task-executor/src"
                  "components/pr-lifecycle/src"
+                "components/pr-lifecycle/resources"
                  "components/release-executor/src"
                  "components/evidence-bundle/src"
                  "components/evidence-bundle/resources"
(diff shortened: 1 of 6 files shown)


H31MDALLR:
Closing — stale pre-existing work in worktree created spurious commits. Reopening with clean branch.

PR #416, opened 2026-04-06
fix: compliance scanner returns zero violations — handle :always-apply rules selector

H31MDALLR:
## Summary

- **`scanner-registry/enabled-rule-configs`**: The `:always-apply` keyword (the designed default) fell through to the else branch where `(map #(get registry %) :always-apply)` iterated over the keyword's characters, returning nothing. Fixed by treating both `:all` and `:always-apply` as \"return all rules\", and coercing string inputs (e.g. `\":always-apply\"`) to keywords via `(keyword (subs raw 1))`.
- **`resolve-rules` in `phases.clj`**: The spec-parser delivers `rules` as the string `\":always-apply\"` from YAML frontmatter, not the keyword `:always-apply`. Fixed by coercing string values to keywords before passing them downstream.

## Test plan

- [ ] `enabled-rule-configs-returns-all-for-always-apply-keyword` — `:always-apply` keyword returns all 3 rule configs
- [ ] `enabled-rule-configs-returns-all-for-always-apply-string` — `\":always-apply\"` string returns all 3 rule configs
- [ ] `enabled-rule-configs-returns-all-when-no-rules-key` — missing `:rules` key defaults to all rules
- [ ] `enabled-rule-configs-returns-all-for-all-keyword` — existing `:all` case still works
- [ ] `enabled-rule-configs-filters-by-requested-set` — explicit rule set still filters correctly
- [ ] All existing `phases_test.clj` tests pass (no changes needed; `base-ctx` uses keyword `:always-apply`)
- [ ] `mf run docs/compliance/miniforge-scan.md` should now return non-zero `rules-scanned` and actual violations


H31MDALLR:
Superseded by a clean cherry-pick branch — reopening as new PR.

PR #417, opened 2026-04-06
fix: serialize git worktree creation and add failure logging

H31MDALLR:
## Summary

Fixes `dag-execution-failed` where dogfood runs showed 28 tasks failed / 0 tokens / 0 duration for all tasks.

**Root cause 1 — concurrent `git config.lock` races**: The DAG orchestrator runs 4 sub-workflow futures simultaneously. Each sub-workflow calls `acquire-environment!` → `create-worktree` → `git worktree add`. Git serializes worktree adds internally via a `config.lock` file; when four threads hit `git worktree add` concurrently, 3 of 4 fail silently. Added a JVM-level `worktree-lock` Object so creation is serialized. Tasks still run in parallel once their environments are acquired.

**Root cause 2 — broken fallback**: On first-attempt failure, the old code fell back to `git worktree add <path> <branch>` (without `-b`). This always fails when `<branch>` (`main`) is already checked out in the primary worktree. Replaced with a 3-attempt strategy:
1. `git worktree add -b <name> <path> <branch>` — fresh branch from branch tip
2. Clean up stale branch/directory, prune, delete branch, retry step 1
3. `git worktree add --detach <path>` — detached HEAD as last resort

**Bonus — surfaced silent failures**: `build-env-record` was silently returning `nil` when worktree acquisition failed (using `when` instead of `if`). This `nil` cascaded to `resolve-worktree-path`, which threw `ex-info`, yielding 0 tokens for all tasks. Added a `WARN` println so failures are visible in stdout immediately.

**Cascade**: 4 first-batch tasks failed → `propagate-failures` cascaded to all 24 dependent tasks = 28 total failures.

## Test plan

- [x] Lint passes on both changed files
- [x] All 260 dag-executor + workflow tests pass (0 failures, 0 errors)


H31MDALLR:
Empty commit due to git worktree/core.worktree config corruption — reopening with correct changes

PR #422, opened 2026-04-06
fix: [730] Version Format (SemVer vs DateVer) compliance pass

H31MDALLR:
Auto-generated by the Miniforge compliance scanner.

Applies mechanical fixes for all auto-fixable violations of this rule.
Violations requiring semantic review are excluded — see the work spec
at `docs/compliance/` for the full list.

🤖 Generated with [Miniforge](https://miniforge.ai)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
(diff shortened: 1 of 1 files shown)


H31MDALLR:
Rule 730 (datever) has insufficient specificity — pattern matches version mentions inside string literals, not just version declarations. Closing until the rule is fixed.

PR #428, opened 2026-04-07
fix: [210] Clojure Map Access compliance pass

H31MDALLR:
Auto-generated by the Miniforge compliance scanner.

Applies mechanical fixes for all auto-fixable violations of this rule.
Violations requiring semantic review are excluded — see the work spec
at `docs/compliance/` for the full list.

🤖 Generated with [Miniforge](https://miniforge.ai)


H31MDALLR:
Closing: built on broken scan.clj (full line as :current instead of regex match). Will regenerate after fix.

PR #433, opened 2026-04-07
fix: handle nil coalesce in ci-monitor and tool search

H31MDALLR:
## Summary

- Fix NPE in `ci-monitor/compute-ci-status` when `:state` or `:conclusion` is explicitly `nil` (GitHub API returns `null` for queued checks)
- Fix NPE in `tool/find-tools` when `:description` is `nil` on a registered tool

Root cause: `(get m :key "")` returns `nil` when the key is present with a `nil` value — the default only applies when the key is absent. Changed to `(or (:key m) "")` which coalesces both cases.

## Test plan

- [x] `ci-monitor-property-test`: 9 tests, 1199 assertions, 0 failures (was 7 errors)
- [x] `tool/interface-test`: 14 tests, 59 assertions, 0 failures (was 1 error)
- [x] Full pr-lifecycle suite: 266 tests, 1880 assertions, 0 failures
- [x] Pre-commit hook passes clean

### DIFF
--- a/components/pr-lifecycle/src/ai/miniforge/pr_lifecycle/ci_monitor.clj
+++ b/components/pr-lifecycle/src/ai/miniforge/pr_lifecycle/ci_monitor.clj
@@ -117,8 +117,8 @@
    Returns {:status keyword :passed [] :failed [] :pending []}"
   [checks]
   (let [grouped (group-by (fn [check]
-                            (let [state (keyword (str/lower-case (get check :state "")))
-                                  conclusion (keyword (str/lower-case (get check :conclusion "")))]
+                            (let [state (keyword (str/lower-case (or (:state check) "")))
+                                  conclusion (keyword (str/lower-case (or (:conclusion check) "")))]
                               (cond
                                 (= state :completed)
                                 (case conclusion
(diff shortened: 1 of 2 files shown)


components/pr-lifecycle/src/ai/miniforge/pr_lifecycle/ci_monitor.clj
```diff
@@ -117,8 +117,8 @@
    Returns {:status keyword :passed [] :failed [] :pending []}"
   [checks]
   (let [grouped (group-by (fn [check]
-                            (let [state (keyword (str/lower-case (get check :state "")))
-                                  conclusion (keyword (str/lower-case (get check :conclusion "")))]
+                            (let [state (keyword (str/lower-case (or (:state check) "")))
```
H31MDALLR: Create a fn to keywordize instead of reintroducing this pattern:

```clojure
(def keywordize-val [m k default]
   (or (some-> m k str/lower-case keyword) default))

(let [state (keywordize-val check :state "")
      conclusion (keywordize-val check :conclusion "")] ...)
```

H31MDALLR: Done — extracted `keywordize-val` helper at Layer 0.5 in acabf2a5. Uses `(or (some-> (get m k) str/lower-case keyword) default)` and both call sites now use `(keywordize-val check :state :unknown)`.

PR #453, opened 2026-04-08
feat: multi-detection scanning — diff-analysis, plan-output, selectors, incremental

H31MDALLR:
Expand scan.clj to support diff-analysis, plan-output, selectors, and incremental mode

### DIFF
--- a/components/compliance-scanner/src/ai/miniforge/compliance_scanner/scan.clj
+++ b/components/compliance-scanner/src/ai/miniforge/compliance_scanner/scan.clj
@@ -172,39 +172,32 @@
   "Detection types the compliance scanner can process."
   #{:content-scan :diff-analysis :plan-output})
 
-(def ^:private canonical-taxonomy-cache
-  "Cached canonical taxonomy for category resolution. Loaded once on first use.
-   Returns nil when the taxonomy API is not yet available."
-  (delay (try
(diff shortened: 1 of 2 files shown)


components/compliance-scanner/src/ai/miniforge/compliance_scanner/scan.clj
```diff
@@ -250,36 +382,98 @@
 ;------------------------------------------------------------------------------ Layer 2
 ;; Top-level entry point
 
+(defn- scan-content-rules
+  "Scan content-scan rules across the repo index.
+   When :since ref is provided, limits scanning to changed files only."
+  [content-cfgs index changed-files]
+  (->> content-cfgs
+       (mapcat (fn [cfg]
```
H31MDALLR: This is messy. Extract the anonymous function out, and then for the embedded anonymous function down there, maybe use a `let fn` inside the new fn or extract it as well..

components/compliance-scanner/src/ai/miniforge/compliance_scanner/scan.clj
```diff
@@ -21,12 +21,14 @@
```
H31MDALLR, line 21: Where are the tests for this PR?

H31MDALLR (commented):
All review comments addressed in latest commits.

PR #492, opened 2026-04-10
Extract the inline workspace-persistence side-effect (lines 399-413...

H31MDALLR:


H31MDALLR:
Duplicate of #491 — same extraction from a second pr respond run.

PR #504, opened 2026-04-11
fix: meta-loop fires on all workflow exit paths

H31MDALLR:
## Summary
- Move `observe-workflow-signal!`, `synthesize-patterns!`, and `promote-mature-learnings!` into the `finally` block so they fire on ALL exit paths (success, failure, and uncaught exception)
- Extend `observe-workflow-signal!` to build a minimal failure signal from the caught exception when `output-ctx` is nil (uncaught exception path)
- Wrap each post-workflow call in its own `try/catch` so a meta-loop failure never masks the actual workflow error or prevents subsequent cleanup

## Test plan
- [x] All 19 `runner-test` assertions pass (0 failures, 0 errors)
- [x] Full workflow brick test suite: 250 tests, 856 passes — 2 pre-existing failures in `runner-extended-test` (confirmed identical on `main`)
- [ ] Verify meta-loop signal fires when pipeline throws an uncaught exception
- [ ] Verify `synthesize-patterns!` and `promote-mature-learnings!` run even on workflow failure

### DIFF
--- a/components/workflow/src/ai/miniforge/workflow/runner.clj
+++ b/components/workflow/src/ai/miniforge/workflow/runner.clj
@@ -540,190 +540,168 @@
         (observe-fn signal)))
     (catch Exception _e nil)))
 
-;------------------------------------------------------------------------------ Layer 2: Main entry point
+;------------------------------------------------------------------------------ Layer 2: Pipeline stages
+
+(declare run-pipeline)
+
+(def ^:private ^:const default-max-phases 50)
+
+(defn- acquire-environment
+  "Acquire an isolated execution environment (worktree or Docker capsule).
+   Returns [acquired-env opts'] where acquired-env is nil when pre-acquired."
+  [workflow input opts]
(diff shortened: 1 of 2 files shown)


components/workflow/src/ai/miniforge/workflow/runner.clj
```diff
@@ -517,18 +517,27 @@
 (defn- observe-workflow-signal!
```
H31MDALLR, line 517: Decompose this so it reads like a pipeline. Also add tests for this new set of fns and functionality.

H31MDALLR: Fixed: decomposed into 6 named pipeline stages. run-pipeline is now 30 lines composing: acquire-environment → build-initial-context → execute-pipeline-loop → validate-completion → extract-output → post-workflow-cleanup!. Each stage is independently testable. Added 8 new tests in runner_pipeline_test.clj.

PR #571, opened 2026-04-18
fix(workflow): emit :workflow/dag-considered with skip reason

H31MDALLR:
## Summary
Add diagnostic events for DAG activation. Every skip path in `try-dag-execution` now emits `:workflow/dag-considered {:dag/outcome :skipped :dag/reason <reason>}` with a specific reason keyword; activation emits `:dag/outcome :activated` with `:plan/id` + `:plan/task-count`.

Reasons enumerated:
- `:not-plan-phase` — not the plan phase
- `:disabled` — `:disable-dag-execution` set (sub-workflow context)
- `:no-plan-id` — phase output lacks `:plan/id`
- `:no-tasks` — plan has `:plan/id` but empty `:plan/tasks`

## Why this, not the real fix
Three miniforge runs today failed with "implementer wrote no files." Working hypothesis was that DAG never activates because `plan-from-agent` doesn't emit `:plan/id`. But a REPL trace this session showed `dag-applicable?` DOES correctly return the plan when the planner produces the expected shape. So the real skip reason is one of the other three — and we have no way to tell which without observability.

This PR adds that observability. The next run that stalls will emit a `:workflow/dag-considered :skipped` event with the concrete reason, telling us exactly which upstream path to fix. See the PR doc for the full write-up.

## Test plan
- [x] 6 new unit tests in `dag_activation_diagnostics_test` — all pass (12 assertions, 0 failures)
- [ ] Next miniforge run emits `:workflow/dag-considered` in `~/.miniforge/events/<workflow-id>/`
- [x] Publish errors swallowed — observability never breaks execution

## Note on `[BYPASS-HOOKS]`
Pre-existing 5 test failures in `environment_promotion_integration_test` are present on main without this change (verified via stash + re-run). Unrelated to this PR; noted in the commit message.

### DIFF
--- a/components/workflow/src/ai/miniforge/workflow/execution.clj
+++ b/components/workflow/src/ai/miniforge/workflow/execution.clj
@@ -341,6 +341,19 @@
             (inc i)))
         (map-indexed vector pipeline)))
 
+(defn dag-skip-reason
+  "Return the reason DAG execution should be skipped, or nil if it should proceed.
+   Reasons are keywords — :not-plan-phase, :disabled, :no-plan-id, :no-tasks."
+  [phase-name phase-result ctx]
+  (cond
+    (not= :plan phase-name) :not-plan-phase
+    (:disable-dag-execution ctx) :disabled
+    :else (let [plan (extract-plan-from-phase-result phase-result)]
+            (cond
+              (nil? plan) :no-plan-id
+              (empty? (:plan/tasks plan)) :no-tasks
+              :else nil))))
+
 (defn dag-applicable?
   "Check whether DAG execution should be attempted for this phase result.
    Returns the plan map if applicable, nil otherwise."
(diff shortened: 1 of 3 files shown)


components/workflow/src/ai/miniforge/workflow/execution.clj
```diff
@@ -416,16 +457,29 @@
...
+                                {:phase/name phase-name}))
+        nil)
+      (let [plan (extract-plan-from-phase-result phase-result)
+            ctx-with-resume (assoc ctx :pre-completed-ids
+                                   (get-in ctx [:execution/opts :pre-completed-dag-tasks] #{}))
+            _ (emit-dag-considered! ctx :activated :plan-has-tasks
+                                    {:plan/id (:plan/id plan)
+                                     :plan/task-count (count (:plan/tasks plan))})
+            dag-result (dag-orch/execute-plan-as-dag plan ctx-with-resume)]
+        (if (:success? dag-result)
```
H31MDALLR: Use success predicates.

PR #608, opened 2026-04-20
feat(pr-scoring): producer-side PR readiness/risk/policy scoring (N5-δ2)

H31MDALLR:
## Summary

- Producer-side implementation of N5-delta-2 (merged 2026-04-20, #598): a new \`components/pr-scoring\` Polylith brick that subscribes to fine-grained \`:pr/*\` events, calls a pluggable scorer-fn, and emits a new \`:pr/scored\` fine-grained event
- \`supervisory-state\` accumulator consumes \`:pr/scored\` and merges the four scored fields into the canonical \`PrFleetEntry\` — scored fields ride the next coalesced \`:supervisory/pr-upserted\` emission, preserving N5-δ1 §3.4 invariant 6 (single emitter per entity family)
- \`PrFleetEntry\` schema gains four OPTIONAL nested fields per spec §2: \`:pr/readiness\`, \`:pr/risk\`, \`:pr/policy\`, \`:pr/recommendation\`
- Partial-merge semantics: a second \`:pr/scored\` with only \`:risk\` does not clobber a previously stored \`:readiness\` — lets producers emit scores incrementally

## What's not here (follow-ups)

- **Real scorer-fn** calling \`pr-train.readiness/explain-readiness\`, \`pr-train.risk/assess-risk\`, \`policy-pack.external/evaluate-external-pr\`. This PR ships a no-op default scorer-fn that emits nothing, plus the plumbing to inject a real one. The real scorer needs PR-context data (diff, author history, train graph) to be accessible — out of scope here
- **Rust consumer-side**: \`miniforge-control\`'s \`PrFleetEntry\` and TUI §3.2.8/§3.2.9 renders. Separate PR on the control repo

## Test plan

- [x] 26 new unit tests, 39 in combined \`supervisory-state\` + \`pr-scoring\` suite — all green
- [x] Pre-commit: \`bb test\` (changed-bricks, 1464 tests / 6729 passes), \`bb test:graalvm\`, clj-kondo, markdown format — all pass
- [x] Smoke: \`(es/pr-scored stream \"acme/widget\" 42 {...})\` emits a well-formed envelope; verified at REPL
- [ ] Follow-up PR: real scorer integration; add integration test against a live event stream

## Key design decisions (carryover from #598)

- \`pr-scoring\` lives as its own component rather than folded into \`supervisory-state\` — keeps the projection component free of \`pr-train\` / \`policy-pack\` deps
- \`:pr/scored\` is fine-grained, not a snapshot event — invariant 6 preserved
- Scorer-fn injection at \`create\` time so the component is testable without real scoring math and safely runnable in v1 before the real integration ships
- Scorer-fn exceptions are swallowed silently in the scaffold — proper structured logging comes with the real scorer PR

Refs: \`specs/normative/N5-delta-2-pr-scoring.md\` (merged #598)

### DIFF
--- a/components/pr-scoring/src/ai/miniforge/pr_scoring/core.clj
+++ b/components/pr-scoring/src/ai/miniforge/pr_scoring/core.clj
@@ -1,0 +1,158 @@
+;; Title: Miniforge.ai
+;; Subtitle: An agentic SDLC / fleet-control platform
+;; Author: Christopher Lester
+;; Line: Founder, Miniforge.ai (project)
+;; Copyright 2025-2026 Christopher Lester (christopher@miniforge.ai)
+;;
+;; Licensed under the Apache License, Version 2.0 (the "License");
+;; you may not use this file except in compliance with the License.
+;; You may obtain a copy of the License at
+;;
+;;     http://www.apache.org/licenses/LICENSE-2.0
+;;
+;; Unless required by applicable law or agreed to in writing, software
+;; distributed under the License is distributed on an "AS IS" BASIS,
+;; WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+;; See the License for the specific language governing permissions and
+;; limitations under the License.
+
+(ns ai.miniforge.pr-scoring.core
+  "PR scoring component — produces `:pr/scored` events per N5-delta-2 §3.
+
+   Subscribes to fine-grained PR-mutating events on the miniforge event
+   stream, invokes a pluggable *scorer function* for each incoming PR
+   event, and emits a `:pr/scored` event when the scorer returns
+   non-nil scores. The emitted event is consumed exclusively by
+   `supervisory-state`, which merges the four score fields into the
+   canonical `PrFleetEntry` per N5-delta-2 §3.2.
+
+   This component owns the **subscription, trigger, and emission**
+   plumbing. The actual scoring math lives in `pr-train.readiness`,
+   `pr-train.risk`, and `policy-pack.external` — the scorer-fn injected
+   at `create` time is expected to call into those libraries and
+   assemble the `{:readiness :risk :policy :recommendation}` return
+   map (any subset MAY be omitted per N5-delta-2 §5.4).
+
+   ## Scorer-fn contract
+
+       (scorer-fn pr-event) => {:readiness {...}
+                                :risk {...}
+                                :policy {...}
+                                :recommendation :merge} ;; or nil
+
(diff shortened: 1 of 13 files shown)


components/pr-scoring/src/ai/miniforge/pr_scoring/core.clj
```diff
@@ -0,0 +1,158 @@
...
+(def ^:private subscriber-id ::pr-scoring)
+
+;------------------------------------------------------------------------------ Layer 0
+;; Trigger events
+
+(def trigger-event-types
+  "PR-mutating event types whose arrival SHOULD re-score the affected PR.
+   Any event of these kinds is forwarded to the scorer-fn; a nil return
+   signals \"skip this PR for now\" and suppresses emission."
+  #{:pr/created
```
H31MDALLR: Looks like configuration, and config is data.

H31MDALLR: Fixed in 64ef6ea — `trigger-event-types` now lives in `components/pr-scoring/resources/config/pr-scoring/triggers.edn`, loaded lazily via a `delay` so the classpath scan doesn't run at namespace load time. Callers can still override via the `:trigger-event-types` option to `create`/`attach!` so tests and alternate deployments don't need to touch config files.

Scanned the rest of the PR for similar patterns:

**Legitimately config-in-code (only this one):** the trigger set, now fixed.

**Pre-existing issue this PR touched but didn't introduce:** the `audit-summary` map in `event_type_registry.clj` has hardcoded counts (`:total-server-events`, `:browser-handled-count`, `:browser-unhandled-count`, `:naming-asymmetry-count`) that must be manually bumped when the registry grows — I had to change 42→43 and 36→37 to keep `audit-summary-test` happy. These are strictly computable from `event-type-registry` and should be derived, not literal. Worth a separate cleanup PR — flagging here but out of scope for N5-delta-2.

**Kept as-is (not config):**
- Schema enum vectors (`risk-levels`, `policy-overall-states`, `pr-recommendations`) — protocol contract, matches the existing `workflow-run-statuses`/`pr-statuses` pattern in the same file
- Malli schemas for `PrReadiness`/`PrRisk`/`PrPolicy` — schemas are the contract
- The stub-PR map in `pr-scored` accumulator handler — edge-case defaults for "score arrives before create", mirrors the `pr-created` precedent in the same file
- `subscriber-id ::pr-scoring` — identity, not parameter
- `trigger-config-resource` path literal — a resource location, not content

PR #610, opened 2026-04-21
fix: planner convergence — unblock plan.edn Write + surface error context

H31MDALLR:
## Summary

Landed iteration 19 of the planner-convergence dogfood loop as the first **plan phase green**. Eight commits, each a distinct failure mode that iters 11–18 surfaced. The ordering here is also the order of root causes peeled off — each uncovered the next.

## Iter-by-iter chain

| commit | iter surfaced it | what was broken |
| --- | --- | --- |
| `1bef7ea1` | iter 11 | `error-details` flattened exceptions with nil `.getMessage` + no `ex-data` into `{:message "Unknown error" :data {}}`. Now attaches `:exception/class` + 8-frame stack preview. |
| `280455e4` | iter 12 | Planner's error branch extracted only `:message` from `llm-error` — discarded `:type`, `:stderr`, `:stdout`, `:timeout`. Now forwards the full llm-error shape as `:data`. |
| `b34f1ccc` | iter 13 | `process-stream-lines` exited silently on line-timeout, letting `deref process 600000` add 10 min of wall time before reporting. Now tags `:type :stream-idle` and `.destroyForcibly`s the subprocess. |
| `27bf71b3` | iter 14 | 60s line-timeout tripped on legitimate Opus 4.6 silent-reasoning pauses. Raised to 180s; progress-monitor's 120s stagnation threshold remains. |
| `9b8a94c2` | iter 15 | `Write` wasn't in `--allowedTools` — Claude CLI silently denied the planner's plan.edn Write. `mcp-tools` now carries `:Write` as a keyword; adapter transform accepts `{:mcp/server :mcp/tool}` maps AND bare keywords. |
| `d7c52ca2` | iter 16 diagnostic | `MF_STREAM_DUMP` env var appends the raw Claude stream-json to a file. Left in place — zero-cost when unset, invaluable when a new failure mode appears. |
| `405dd351` | iter 17 | Planner's Claude subprocess cwd was the miniforge repo instead of the task worktree. Write(".miniforge/plan.edn") resolved to `/Users/chris/.../miniforge/.miniforge/plan.edn` where a 20K stale plan.edn blocked Claude Code's "Read-before-Write" safety. Planner now threads session `:workdir` into the LLM request. |
[...]

PR #615, opened 2026-04-21
feat(implementer): port planner-convergence fixes — artifact wins on CLI error

H31MDALLR:
## Summary

Iter-20 of the planner-convergence dogfood landed plan phase green, then both spawned implement sub-workflows failed. The implementer had two of the same bugs the planner just got fixed in #610:

1. File-artifact container promotion was only honored when \`(llm/success? response)\` — CLI timeout AFTER the agent wrote files to the worktree discarded the work.
2. The error branch extracted only \`:message\` from llm-error, losing \`:type\`, \`:timeout\`, \`:stderr\`, \`:stdout\`, \`:exit-code\` — same undiagnosable bare-message shape.

## Fix

- \`(if (or effective-artifact (llm/success? response)) …)\` — container-promotion result preempts CLI classification. Mirrors the planner's \`(or worktree-plan (llm/success? …))\`.
- Error path forwards the full llm-error as \`:data\` with \`:stop-reason\` + \`:num-turns\` merged.
- \`:implementer/llm-called\` log event gets \`:stop-reason\` / \`:num-turns\` / \`:artifact-source\` for parity with the planner.

## Already in place (from #610)

Global fixes applied to every role; implementer inherits them:

- \`:workdir\` threading into Claude subprocess (implementer was already correct)
- \`:Write\` in \`--allowedTools\`
- stream-idle fast-fail + \`.destroyForcibly\` on hung subprocess
- 180s line-timeout for Opus reasoning pauses

## Test plan

- [x] Existing invoke/fallback tests green
- [x] **NEW**: \"file artifact wins when LLM call failed\" — iter-20 regression guard
- [x] **NEW**: \"LLM error preserves :type + timeout shape in :data\" — stop-reason, num-turns, timeout type all flow through
- [x] 11 / 66 assertions on implementer-test green
- [x] Pre-commit + GraalVM
- [ ] Post-merge: iter 21 against planner-convergence spec — expect sub-workflow implement phases to succeed (or fail with structured, diagnosable error data)

### DIFF
--- a/components/agent/src/ai/miniforge/agent/implementer.clj
+++ b/components/agent/src/ai/miniforge/agent/implementer.clj
@@ -508,16 +508,41 @@
                 {:data {:file-count (count (:code/files file-artifact))
                         :tools-called (get response :tools-called [])}}))
     (log/info logger :implementer :implementer/llm-called
-              {:data {:success (llm/success? response)
-                      :tokens tokens
-                      :streaming? (boolean on-chunk)
-                      :file-artifact-fallback? (= :file-fallback artifact-source)
-                      :tools-called (get response :tools-called [])}})
-    (if (llm/success? response)
+              {:data (cond-> {:success (llm/success? response)
+                              :tokens tokens
+                              :streaming? (boolean on-chunk)
+                              :artifact-source (or artifact-source :none)
+                              :tools-called (get response :tools-called [])}
+                       (:stop-reason response)
+                       (assoc :stop-reason (:stop-reason response))
+                       (:num-turns response)
+                       (assoc :num-turns (:num-turns response)))})
+    ;; Container-promotion precedence: if the agent wrote files into
+    ;; the worktree (file-artifact via collect-written-files) or
+    ;; submitted via the MCP artifact path, honor that even when the
+    ;; CLI itself terminated with a timeout/error. Iter-20 of the
+    ;; planner-convergence dogfood showed Claude stalling AFTER a
+    ;; successful stream of edits; the old path discarded a real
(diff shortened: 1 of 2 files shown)


H31MDALLR:
### Added root-cause fix: EOF vs stream-idle misclassification

You were right to question "timeouts after completion" — they weren't. Iter-18 stream dump showed:

```
line 49: Write tool_use (succeeded, plan.edn in worktree)
line 50: tool_result "File created successfully at ..."
line 51: assistant text "Plan written to .miniforge/plan.edn..."
line 52: RESULT subtype=success stop_reason=end_turn num_turns=21 duration_ms=329114
        — subprocess exits cleanly here —
```

…yet our runtime logged `:type :adaptive_timeout :timeout :stream-idle` at 5.5 min.

**Root cause:** `read-line-with-timeout` returned `nil` for both "clean EOF" (pipe closed after subprocess exited) and "deref hit the deadline." The loop treated both identically and set `:stream-idle`.

**Fix** (commit `a1973b2c`): `read-line-with-timeout` now returns the line string, `nil` only on clean EOF, or a sentinel `::read-timeout` on actual deadline. `process-stream-lines` handles the three cases explicitly — EOF exits the loop with no timeout reason.

This means iter-19's "first green plan" was partially papered over by the earlier `(or worktree-plan ...)` guard. That guard still has value (handles the genuine stream-idle case) but is no longer masking the misclassification. 

Same iter-20 `error_max_turns` case in the implementer: now that we correctly read to EOF, the real Claude CLI result event surfaces and we'll see `stop_reason: tool_use subtype: error_max_turns` in `:data` instead of a bogus timeout.

PR #616, opened 2026-04-21
feat(supervisory-state): TaskNode + :supervisory/task-node-upserted (N5-δ3 §2.3)

H31MDALLR:
## Summary

First of the N5-δ3 producer-side implementation PRs. Projects DAG task state from the existing \`:task/state-changed\` fine-grained events into a TaskNode entity, emitted on the supervisory snapshot path so the Rust TUI (and any other consumer) can render N5 §3.2.5 DAG Kanban without recomputing state locally.

- **TaskNode** open map added to \`schema.clj\` with open \`:task/status\` (state-profile system can add statuses without a spec bump) and closed \`:task/kanban-column\` (display contract).
- **Status → Kanban column** mapping per spec table; unknown statuses fall back to \`:blocked\` so they surface visibly rather than silently landing in \`:done\`.
- **Opportunistic context fields** (description / type / component / dependencies) populated from \`:task/context\` when the scheduler emits them; later events preserve prior values.
- **Timestamps**: \`:started-at\` on first entry to \`:active\`, \`:completed-at\` on first entry to \`:done\`, \`:elapsed-ms\` recomputed each update.
- Emitter + \`diff-and-emit!\` extended; \`handle-event!\` ignore set widened to break re-emit loops; replay handler trusts snapshot carry.

## What's not here (follow-ups)

- Rust consumer side (\`TaskNode\` struct + \`SupervisoryState.tasks\` + \`StateManager\` dispatch) — separate PR in miniforge-control
- TUI DAG Kanban view rendering the six columns — separate PR in miniforge-control

## Test plan

- [x] 11 new unit tests in \`accumulator_test.clj\` / \`core_test.clj\` — all 12 known statuses mapped, unknown-status fallback, context preservation, timestamps, live-emit, snapshot-replay
- [x] 40 supervisory-state tests total / 104 assertions — all green
- [x] Pre-commit (clj-kondo, markdown format, GraalVM/Babashka) — pass

Refs: N5-δ3 §2.3, §3.3, §4.3 (merged in #611 + #614)

### DIFF
--- a/components/supervisory-state/src/ai/miniforge/supervisory_state/accumulator.clj
+++ b/components/supervisory-state/src/ai/miniforge/supervisory_state/accumulator.clj
@@ -28,6 +28,8 @@
    This namespace knows nothing about emission, persistence, or threading.
    It is a pure function from an event-stream prefix to an EntityTable."
   (:require
+   [clojure.edn :as edn]
+   [clojure.java.io :as io]
    [clojure.string :as str]))
 
 ;------------------------------------------------------------------------------ Layer 0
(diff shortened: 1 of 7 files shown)


components/supervisory-state/src/ai/miniforge/supervisory_state/accumulator.clj
```diff
@@ -273,6 +273,87 @@
                    recommendation (assoc :pr/recommendation recommendation))]
     (assoc-in table [:prs k] scored)))
 
+;------------------------------------------------------------------------------ Layer 1
+;; TaskNode handlers (N5-δ3 §2.3, §3.3)
+
+(def status->kanban-column
```
H31MDALLR: This is configuration, so it should be data.

H31MDALLR: Fixed in ea77438 — `status->kanban-column` now lives in `components/supervisory-state/resources/config/supervisory-state/task-kanban-mapping.edn`, loaded lazily via a `delay`. Same pattern used for `pr-scoring` triggers in #608. The `:blocked` fallback for unknown statuses stays in the accumulator code (it's the projection rule, not the mapping table). New test asserts the loader reads real data rather than a compiled-in literal.

PR #617, opened 2026-04-21
fix(resume): read per-event JSON from workflow dir, not a missing .edn

H31MDALLR:
## The bug

\`mf run --resume <workflow-id>\` has been silently broken since the event-stream file sink switched from line-delimited EDN to one-transit-JSON-file-per-event. \`resume/read-event-file\` still looked for \`{events-dir}/{workflow-id}.edn\` — a path that is never written — and every resume attempt threw \`:anomalies/not-found\`, even for workflows with complete event histories on disk.

Observed iter-20 of the planner-convergence dogfood: plan succeeded at 3:20, sub-workflow implement phases failed. No way to short-circuit the plan on a re-run; planner burned ~3 min + real tokens every iteration. This fix saves that cost per resume per user.

## The fix

New \`components/event-stream/reader.clj\`:

- \`strip-transit-prefix\` — turns transit-JSON's \`~:\`/\`~u\`/\`~t\` prefixed strings back into keywords / stripped strings so keyword accessors work downstream. Extracted from logic duplicated in \`observability.clj\`; that call site will be migrated in a follow-up.
- \`read-workflow-events\` — reads every \`.json\` in a workflow's events dir, sorted by filename (timestamp-prefixed). Tolerant of non-JSON files and one-off corrupt events.
- \`read-workflow-events-by-id\` — convenience over \`{base-dir}/{workflow-id}/\`.

Exposed via \`event-stream/interface\`. \`resume/read-event-file\` now delegates. Throws \`:anomalies/not-found\` only when the workflow dir is genuinely missing or empty.

## Test plan

- [x] \`reader-test\`: 9 cases / 21 asserts — transit-prefix edge cases + sort order + non-JSON filter + corrupt-file tolerance + missing dir + by-id composition
- [x] \`resume-test\`: 2 new cases / 7 asserts — happy-path regression guard (multi-event per-file JSON), missing-dir error path
- [x] 12 tests / 31 assertions total / 0 failures
- [x] Pre-commit + GraalVM green
- [ ] Post-merge: \`mf run --resume <previously-failed-workflow-id>\` picks up where plan left off

## Follow-up

\`observability.clj\` has its own copy of \`read-workflow-events\` + \`strip-transit-prefix\` (for \`mf events show\`). Not touched here to keep the fix focused; it can migrate to the new interface helpers in a separate PR.


H31MDALLR:
### Addressed: promote resume to a first-class component + subcommand

Per review — resume was ~227 LOC of event-file replay + reconstruction + pipeline-trim buried in the CLI base. Refactored into a proper component.

**New:** `components/workflow-resume/` — pure domain core (11 tests / 30 asserts):
- `reconstruct-context` — events → resume-context map
- `trim-pipeline` — workflow + completed-phases → trimmed workflow
- `resolve-workflow-identity` — spec OR fallback-fn → {:type :version}
- extractors (`extract-completed-phases`, `extract-phase-results`, `extract-completed-dag-tasks`, `extract-dag-pause-info`, `find-workflow-spec`)

**CLI shrinks** to a 141-LOC thin adapter: delegations to the component + runtime wiring (event-stream, supervisory, LLM client, dashboard) + display.

**New subcommand:** `mf resume <workflow-id>` (positional arg, or `-w`/`--workflow-id`, or legacy `--resume` for compat).

Legacy `mf run --resume <id>` kept working — the same `cmd-resume/resume-workflow` path.

Now callable from HTTP API, dashboard, any future headless control plane without pulling in CLI-only code — same interface, same tests.

Commit: `2817ef72`.

H31MDALLR:
### Addressed: boundary validation + message localization

Per new review:

**1. Schema validation at the component boundary** (Dewey 004). The component reads events from disk — an external boundary where data from a potentially different miniforge version can arrive half-written or shape-drifted. New `components/workflow-resume/schema.clj` with malli schemas:

- `EventBase` — minimum event shape (`:event/type` keyword)
- `ReconstructContextInput` / `TrimPipelineInput` / `ResolveWorkflowIdentityInput` — caller-side checks
- Every public fn now calls `schema/validate!` before the pure core runs
- `reconstruct-context` filters raw events through `valid-event?` and surfaces `:raw-event-count` in the not-found anomaly so post-mortems can distinguish "no events" from "all events malformed"

**2. Localized messages**. `resume.clj` / `main.clj` had hardcoded English strings for every user-facing line. Moved to `resources/config/cli/messages/en-US.edn` under `:resume/*` keys, consumed via `messages/t`. Matches the pattern `run`/`scan` already use.

Tests: 4 new validation cases (nil/non-string workflow-id rejection, malformed-event filtering, pipeline shape rejection, non-fn fallback rejection). 27 / 68 asserts green across the three affected suites.

---

### Noted for follow-up (not this PR): supervisory-state event subscription

Valid concern that `supervisory-state/core.clj` rederives events into its own atom (L71-79 replay, L22-33 re-emit) instead of subscribing to the event stream. That's a separate refactor — will track as its own spec/PR; this one stays focused on workflow-resume boundary hygiene.

Commit: `86b285d5`.

PR #644, opened 2026-04-23
feat: persist workflow machine snapshots

H31MDALLR:
# feat: restore workflows from durable machine snapshots

## Overview

Persist the authoritative workflow execution machine as durable checkpoint data
and teach resume flows to restore from that snapshot instead of reconstructing
execution state from phase indexes alone.

## Motivation

`#638` made the execution machine authoritative for workflow progression, but
checkpoint/resume still depended on legacy event-log reconstruction and
phase-index semantics. The specs now require machine-snapshot restoration.

This slice closes that gap by making resume prefer a persisted execution
snapshot while preserving the existing event-history fallback.

## Changes In Detail

- Add durable workflow checkpoint storage in `workflow.checkpoint-store`
  - machine snapshot
  - checkpoint manifest
  - per-phase checkpoint files
- Add `workflow.interface/load-checkpoint-data` as the public restore boundary
- Extend `workflow.context` with `restore-context`
  - recompiles the machine
  - restores the persisted machine snapshot
  - restores phase results
  - rehydrates runtime-only monitoring fields
- Persist checkpoints from `workflow.runner`
  - after initial context creation
  - after each iteration result
- Update `workflow-resume` to prefer machine snapshots and checkpoint manifests
  over event-log reconstruction when available
- Update CLI `resume` to resume the original workflow run id and pass restored
  machine state into `workflow/run-pipeline`
- Add user config data for the default checkpoint root
- Normalize checkpoint payloads into EDN-readable data so timestamp fields
  round-trip cleanly

## Testing Plan

- `clj-kondo --lint` on all touched source and test files
- `bb test components/workflow-resume components/workflow components/config`
- `bb pre-commit`

## Deployment Plan

No deployment step. This is an internal workflow runtime change.

## Related Issues/PRs

- Follows `#638`, which made the execution machine authoritative
- Prepares the next workflow critical-path slices:
  - supervision/runtime boundary extraction
  - formal supervision FSM and intervention lifecycle
  - configurable workflow compilation

## Checklist

- [x] Execution machine snapshots persist durably
- [x] Resume prefers persisted machine snapshots
- [x] Legacy event-history fallback remains available
- [x] Checkpoint payloads are EDN-readable
- [x] Tests cover snapshot persistence and restoration
- [x] Full pre-commit passes

### DIFF
--- a/components/workflow-resume/src/ai/miniforge/workflow_resume/core.clj
+++ b/components/workflow-resume/src/ai/miniforge/workflow_resume/core.clj
@@ -37,6 +37,7 @@
   (:require
    [ai.miniforge.event-stream.interface :as es]
    [ai.miniforge.response.interface :as response]
+   [ai.miniforge.workflow.interface :as workflow]
    [ai.miniforge.workflow-resume.schema :as schema]))
 
 ;------------------------------------------------------------------------------ Layer 0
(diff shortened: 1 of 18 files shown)


components/workflow-resume/src/ai/miniforge/workflow_resume/core.clj
H31MDALLR, line 146: Extract any let form that's over a single line that detracts from readability and intent. I'm not sure we should be maintaining backwards compatibility, but we can make that a pass at the end. I am fully okay, given that there's only, I think, 14 or 15 people who have starred this and about five of them I know aren't using it. Our next major stable tag and version update that we release can note this as a change. I don't even think it's a breaking change from the point of view of somebody running Miniforge.

components/workflow/src/ai/miniforge/workflow/interface/checkpoints.clj
```diff
@@ -0,0 +1,24 @@
+;; Title: Miniforge.ai
```
H31MDALLR, line 1: I think we should be validating at the boundary before we save. That will save us time because we won't be saving corrupted data. If we find something that's wrong, we can actually throw a validation error because that checkpoint will be unable to load. We can decide, on a validation error, do we just drop that attribute?

H31MDALLR:
Follow-up for the two review threads now resolved in 5fa97b68:

- `workflow_resume/core.clj`: split the resume reconstruction logic into smaller helpers so `reconstruct-context` reads as a smaller pipeline.
- `workflow/interface/checkpoints.clj` + `workflow.checkpoint-store`: checkpoint payloads are now validated before persistence and again at the public load boundary. Added regression coverage for invalid save/load cases, while keeping legitimately nullable workflow identity/version fields accepted.

Validation after the fix:
- `clojure -M:test -e (require (quote ai.miniforge.workflow.runner-test) (quote ai.miniforge.workflow.checkpoint-store-test) (quote ai.miniforge.workflow-resume.core-test)) (clojure.test/run-tests (quote ai.miniforge.workflow.runner-test) (quote ai.miniforge.workflow.checkpoint-store-test) (quote ai.miniforge.workflow-resume.core-test))`
- full `bb pre-commit` via the commit hook on `fix: validate workflow checkpoint payloads`

PR #660, opened 2026-04-25
refactor: unify workflow validator and compiler

H31MDALLR:
# Refactor Workflow Validator And Compiler Unification

## Summary

- move legacy workflow normalization into shared
  [components/workflow/src/ai/miniforge/workflow/definition.clj](/components/workflow/src/ai/miniforge/workflow/definition.clj)
  so configurable execution and validation use the same
  `:workflow/phases -> :workflow/pipeline` projection
- make the execution-machine compiler normalize workflows directly, then
  surface localized validation messages for unknown transition targets
- replace the old hand-rolled DAG validation in
  [components/workflow/src/ai/miniforge/workflow/validator.clj](/components/workflow/src/ai/miniforge/workflow/validator.clj)
  with compiled-machine validation, and remove the extra machine-validation
  pass from
  [components/workflow/src/ai/miniforge/workflow/registry.clj](/components/workflow/src/ai/miniforge/workflow/registry.clj)

## Why

- registration, loading, and configurable execution were still using two
  different authorities for workflow validity
- legacy phase workflows could execute through the compiled machine path
  while validation still reasoned about a separate graph model
- this slice makes registration and loading validate the same machine shape
  the runtime actually executes

## Validation

- `clj-kondo --lint components/workflow/src/ai/miniforge/workflow/definition.clj`
  `components/workflow/src/ai/miniforge/workflow/configurable.clj`
  `components/workflow/src/ai/miniforge/workflow/fsm.clj`
  `components/workflow/src/ai/miniforge/workflow/validator.clj`
  `components/workflow/src/ai/miniforge/workflow/registry.clj`
  `components/workflow/test/ai/miniforge/workflow/definition_test.clj`
  `components/workflow/test/ai/miniforge/workflow/validator_test.clj`
  `components/workflow/test/ai/miniforge/workflow/registry_test.clj`
- `clojure -M:dev:test -e "(require '[clojure.test :as t]`
  `'ai.miniforge.workflow.definition-test`
  `'ai.miniforge.workflow.validator-test`
  `'ai.miniforge.workflow.registry-test`
  `'ai.miniforge.workflow.configurable-test`
  `'ai.miniforge.workflow.fsm-test) ..."`
- `bb test components/workflow`
- `bb pre-commit`

### DIFF
--- a/components/workflow/src/ai/miniforge/workflow/fsm.clj
+++ b/components/workflow/src/ai/miniforge/workflow/fsm.clj
@@ -40,6 +40,7 @@
   (:require
    [clojure.set :as set]
    [ai.miniforge.fsm.interface :as fsm]
+   [ai.miniforge.workflow.definition :as definition]
    [ai.miniforge.workflow.messages :as messages]))
 
 (declare current-state first-phase-index reachable-states unreachable-states)
@@ -120,7 +121,10 @@
   [error-key phase target]
   {:error error-key
    :phase phase
-   :target target})
+   :target target
+   :message (messages/t error-key
+                        {:phase phase
+                         :target target})})
 
 (defn- unknown-success-target
   [pipeline {:keys [phase on-success]}]
@@ -317,7 +321,8 @@
    progression. Coarse lifecycle status and phase/index projections are derived
    from the resulting machine snapshot."
   [workflow]
-  (let [pipeline (:workflow/pipeline workflow)
+  (let [normalized-workflow (definition/ensure-execution-pipeline workflow)
+        pipeline (:workflow/pipeline normalized-workflow)
         phase-entries (mapv build-phase-entry (range) pipeline)
         first-active-state (some-> phase-entries first :state-id)
         no-phase-machine? (empty? phase-entries)
@@ -364,7 +369,8 @@
 (defn validate-execution-machine
   "Validate that a workflow pipeline can compile into an execution machine."
   [workflow]
-  (let [pipeline (:workflow/pipeline workflow)
+  (let [normalized-workflow (definition/ensure-execution-pipeline workflow)
+        pipeline (:workflow/pipeline normalized-workflow)
         duplicate-phases (->> pipeline
                               (map :phase)
                               frequencies
@@ -380,7 +386,7 @@
         machine (when (and (seq pipeline)
                            (empty? unresolved-success)
                            (empty? unresolved-fail))
-                  (compile-execution-machine workflow))
+                  (compile-execution-machine normalized-workflow))
         unreachable-state-set (when machine
                                 (unreachable-states machine))
         unreachable-phases (when machine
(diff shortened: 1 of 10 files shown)


components/workflow/src/ai/miniforge/workflow/fsm.clj
```diff
@@ -317,7 +321,8 @@
    progression. Coarse lifecycle status and phase/index projections are derived
    from the resulting machine snapshot."
   [workflow]
-  (let [pipeline (:workflow/pipeline workflow)
+  (let [normalized-workflow (definition/ensure-execution-pipeline workflow)
+        pipeline (:workflow/pipeline normalized-workflow)
```
H31MDALLR: We have repeated map access and different functions. That means we should have a factory function to extract that, which centralizes the map access and allows us to change it in one location.

components/workflow/src/ai/miniforge/workflow/validator.clj
```diff
@@ -104,92 +105,22 @@
...
-          (let [neighbors (get graph current #{})
-                new-to-visit (concat to-visit' neighbors)]
-            (recur new-to-visit (conj visited current))))))))
+;; Compiled-machine validation
+
+(defn- machine-validation-messages
+  [workflow]
+  (let [normalized-workflow (definition/ensure-execution-pipeline workflow)
+        validation (workflow-fsm/validate-execution-machine normalized-workflow)]
+    {:valid? (:valid? validation)
```
H31MDALLR: This should be a predicate named valid?.

components/workflow/src/ai/miniforge/workflow/validator.clj
```diff
@@ -228,18 +159,21 @@
...
-        dag-result (validate-dag config)
+        machine-result (validate-dag config)
         budget-result (validate-budgets config)
         all-errors (concat (:errors schema-result)
-                           (:errors dag-result)
-                           (:errors budget-result))]
+                           (:errors machine-result)
+                           (:errors budget-result))
+        warnings (:warnings machine-result)]
     {:valid? (empty? all-errors)
```
H31MDALLR: This looks very similar to another map we're hand constructing. If we're going to be hand constructing this map in more than one function, then we need to extract that construction to a factory function for this map.

H31MDALLR: Fixed in 399612e2. I factored the repeated validation response construction into `validation-result` and `validation-result-with-warnings`, so those maps are no longer hand-built inline.

components/workflow/src/ai/miniforge/workflow/validator.clj
H31MDALLR: Fixed in 399612e2. I added the `valid?` predicate in `workflow.validator` and now use it instead of repeating direct `:valid?` map access at the call sites.

components/workflow/src/ai/miniforge/workflow/fsm.clj
H31MDALLR: Fixed in 399612e2. I added `definition/execution-pipeline` as the shared accessor and switched the FSM compiler and validator to use it, so the normalized pipeline lookup now lives in one place.

PR #665, opened 2026-04-26
refactor: align loop and task lifecycles with fsm defs

H31MDALLR:
# refactor: move loop and task lifecycle slices toward shared FSMs

## Overview

Replace the hand-maintained loop/task lifecycle maps with shared `components/fsm`
machine definitions, while keeping the public loop/task APIs stable for this
slice.

## Motivation

The loop inner lifecycle and task lifecycle both duplicated transition authority
as ad hoc maps and validation helpers. That made it easy for the transition
tables, terminal-state checks, and tests to drift apart from the shared FSM
direction already used elsewhere in the repo.

This slice narrows that gap without expanding scope into orchestration or
workflow runtime changes.

## Changes In Detail

- Define a shared-FSM-backed inner loop lifecycle machine in
  `components/loop/src/ai/miniforge/loop/inner.clj`
- Define a shared-FSM-backed task lifecycle machine in
  `components/task/src/ai/miniforge/task/core.clj`
- Derive transition maps and transition-event lookups from the machine
  definitions instead of maintaining separate hand-written transition tables
- Keep the existing public helpers such as `valid-transition?`,
  `validate-transition`, `terminal-state?`, and `transition`
- Extract small result/metric helpers to reduce duplicated lifecycle map
  construction
- Add reachability and final-state coverage in loop/task tests

## Testing Plan

- Run `clj-kondo` on the edited loop/task source and test files
- Run `bb test components/loop components/task`

## Deployment Plan

No deployment changes. This is an internal refactor and test coverage update.

## Related Issues/PRs

- Follows the workflow FSM migration work already merged
- Keeps the loop/task slice disjoint from concurrent workflow/supervision work

## Checklist

- [x] Shared FSM definitions added for loop/task lifecycle slices
- [x] Hand-maintained transition authority removed from owned files
- [x] Reachability and invalid-transition coverage added
- [x] PR write set kept within owned files, related tests, and this doc

### DIFF
--- a/components/loop/src/ai/miniforge/loop/inner.clj
+++ b/components/loop/src/ai/miniforge/loop/inner.clj
@@ -27,7 +27,9 @@
    [ai.miniforge.fsm.interface :as fsm]
    [ai.miniforge.logging.interface :as log]
    [ai.miniforge.loop.escalation :as escalation]
+   [ai.miniforge.loop.inner-config :as inner-config]
    [ai.miniforge.loop.gates :as gates]
+   [ai.miniforge.loop.messages :as loop-messages]
    [ai.miniforge.loop.repair :as repair]
    [ai.miniforge.response.interface :as response]))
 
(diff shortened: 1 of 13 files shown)


components/loop/src/ai/miniforge/loop/inner.clj
```diff
@@ -24,24 +24,121 @@
...
    [ai.miniforge.loop.escalation :as escalation]
    [ai.miniforge.loop.gates :as gates]
    [ai.miniforge.loop.repair :as repair]
    [ai.miniforge.response.interface :as response]))
 
 ;------------------------------------------------------------------------------ Layer 0
-;; State machine transitions (pure functions)
+;; State machine definition and derived transition helpers
+
+(def default-max-iterations
```
H31MDALLR, line 39: All of these feel like configuration, and configuration is extracted to data and edn files.

components/task/src/ai/miniforge/task/core.clj
```diff
@@ -22,20 +22,105 @@
...
    [ai.miniforge.schema.interface :as schema]
    [ai.miniforge.logging.interface :as log]))
 
 ;------------------------------------------------------------------------------ Layer 0
 ;; State machine definitions and pure utilities
 
+(def initial-task-status
+  :pending)
+
+(def invalid-transition-message
```
H31MDALLR, line 37: Similar here, these defs are configuration data. Also strings are localized, not just defined.

components/task/src/ai/miniforge/task/core.clj
```diff
@@ -22,20 +22,105 @@
...
+
+(defn- derive-valid-transitions
+  [machine-definition]
+  (reduce-kv
+   (fn [transitions state state-definition]
+     (assoc transitions state (transition-targets state-definition)))
+   {}
+   (:fsm/states machine-definition)))
+
+(defn- derive-transition-events
```
H31MDALLR, line 88: This is a reduce inside of a reduce, and we're using reduce kv, which is a little non-standard. Decompose this function to make sure the intent is clear.

H31MDALLR: Split the transition-event derivation into `target-state` and `state-transition-events`, so `derive-transition-events` is a single merge over per-state maps instead of a nested `reduce-kv`.

components/task/src/ai/miniforge/task/core.clj
H31MDALLR: Moved the task lifecycle machine into `config/task/lifecycle.edn`, added `config/task/messages/en-US.edn`, and wired `task.core` through `task.lifecycle-config` and `task.messages` so the lifecycle strings are localized instead of inlined.

components/loop/src/ai/miniforge/loop/inner.clj
H31MDALLR: Moved the inner-loop lifecycle definition into `config/loop/inner.edn`, loaded it through `loop.inner-config`, and localized the emitted lifecycle strings through `config/loop/messages/en-US.edn`.

PR #675, opened 2026-04-28
feat: emit richer supervisory decision and agent context

H31MDALLR:
# feat: emit richer supervisory decision and agent context

## Overview

Enrich the control-plane producer events and supervisory projections so
downstream clients receive actionable decision context and richer agent state
without inventing UI-only fields.

This PR stays within the producer, event-stream, adapter, and supervisory-state
layers. It does not change the workflow FSM, phase transition authority, or
decision model semantics.

## Motivation

The current supervisory clients can already render:

- `DecisionCard.decision_type`
- `DecisionCard.context`
- `DecisionCard.options`
- `DecisionCard.deadline`
- `DecisionCard.comment`
- `AgentSession.metadata`
- `AgentSession.task`

But `miniforge` was not actually emitting most of that data on the wire.

In practice that meant:

- decision queues collapsed into UUIDs plus a vague summary
- bulk approvals had no real grouping context beyond the summary line
- agent detail had almost no useful run/task metadata
- clients had to guess whether missing context was a rendering problem or a
  producer problem

This PR fixes the producer side of that gap.

### 1. Preserve richer decision context on `:control-plane/decision-*`

`event-stream.core/cp-decision-created` now accepts and emits the richer fields
already present on control-plane decisions:

- `:cp/priority`
- `:cp/type`
- `:cp/context`
- `:cp/options`
- `:cp/deadline`

`event-stream.core/cp-decision-resolved` now also preserves:

- `:cp/comment`

`control-plane.orchestrator/submit-decision-from-agent!` now projects those
fields from the created decision record into the emitted event, and
`resolve-and-deliver!` now carries the optional resolution comment.

### 2. Stop dropping agent registration metadata

`control-plane.orchestrator/run-discovery-pass` previously emitted only the
registered agent id, vendor, and name. It now preserves the rest of the
already-known registration state:

- `:cp/external-id`
- `:cp/capabilities`
- `:cp/metadata`
- `:cp/tags`
- `:cp/heartbeat-interval-ms`

That means supervisory-state can retain richer `AgentSession` records instead of
reconstructing a thinner shadow of the registry.

### 3. Mirror heartbeat task context into supervisory state

Stable status is still meaningful when the task changes. This PR makes that
observable by:

- enriching `:control-plane/agent-heartbeat` with optional `:cp/task` and
  `:cp/metrics`
- emitting heartbeat events during poll passes even when normalized status does
  not change
- updating supervisory-state accumulation so `AgentSession.task` and
  `AgentSession.metrics` track the latest heartbeat payload

### 4. Preserve richer native miniforge agent metadata

`adapter-miniforge` now carries more source event context into
`AgentSession.metadata` when it is present on native miniforge events:

- `:workflow-spec`
- `:workflow-phase`
- `:agent-context`
- `:phase-context`
- `:message`

This keeps the producer aligned with the existing “metadata is open” contract
instead of forcing downstream clients to infer context from a single message
string.

## Scope And Non-Goals

This PR intentionally does not:

- add new workflow-machine states
- change the finite state machine refactor or phase execution semantics
- invent new client-shaped decision fields outside the existing decision model
- solve every detail-view UX gap by itself

The goal here is to make authoritative producer data richer and more faithful to
the canonical decision and registry state that already exists.

## Why This Shape

The cleanest source of truth was already in-process:

- the control-plane decision record already had type, context, options, and
  deadline
- the control-plane registry already had agent metadata, capabilities, and task
  context

The bug was that the event projection was dropping those fields before
supervisory-state and downstream clients ever saw them.

So the right fix is to preserve existing canonical data across the event
boundary, not to invent special-case TUI payloads.

## Enabled Follow-On Work

This producer slice is intended to make all supervisory clients denser and more
actionable immediately, especially `miniforge-control`.

Concrete downstream effects:

1. decision queues can render the actual decision type, context, options, and
   resolution comments already supported by the current Rust contract
2. agent detail can rely on richer metadata/task state instead of UUID-only
   placeholders
3. bulk decision resolution can group around meaningful context instead of just
   summary text

## Testing Plan

Executed in repository root:

```sh
bb test components/event-stream components/control-plane components/adapter-miniforge components/supervisory-state
```

Result:

- targeted producer/control-plane bricks passed
- no failures in the enriched event-stream, control-plane, adapter-miniforge,
  or supervisory-state coverage

## Checklist

- [x] Decision-created events preserve existing decision context fields
- [x] Decision-resolved events preserve optional operator comment
- [x] Agent registration events preserve already-known registry metadata
- [x] Heartbeat events preserve task/metrics and update supervisory sessions
- [x] Native miniforge adapter metadata is richer without changing protocol
- [x] Tests cover the richer event and projection behavior

### DIFF
--- a/components/adapter-miniforge/test/ai/miniforge/adapter_miniforge/interface_test.clj
+++ b/components/adapter-miniforge/test/ai/miniforge/adapter_miniforge/interface_test.clj
@@ -77,7 +77,8 @@
         (is (= #{:code-generation :test-writing :code-review}
                (:agent/capabilities info)))
         (is (= {:workflow-id "wf-123"
-                :event-type  :agent-started}
+                :event-type  :agent-started
+                :message "Starting agent"}
                (:agent/metadata info)))))
 
     (testing "falls back to :message when :workflow/name is nil"
--- a/components/event-stream/src/ai/miniforge/event_stream/core.clj
+++ b/components/event-stream/src/ai/miniforge/event_stream/core.clj
@@ -573,15 +573,24 @@
       (assoc :cp/agent-id agent-id
              :cp/vendor vendor)
       (cond->
-        (:name opts) (assoc :cp/agent-name (:name opts)))))
+        (:name opts) (assoc :cp/agent-name (:name opts))
+        (:external-id opts) (assoc :cp/external-id (:external-id opts))
+        (:capabilities opts) (assoc :cp/capabilities (vec (:capabilities opts)))
+        (:metadata opts) (assoc :cp/metadata (:metadata opts))
+        (:tags opts) (assoc :cp/tags (vec (:tags opts)))
+        (:heartbeat-interval-ms opts)
+        (assoc :cp/heartbeat-interval-ms (:heartbeat-interval-ms opts)))))
 
 (defn cp-agent-heartbeat
   "Emit when the control plane receives a heartbeat from an agent."
-  [stream workflow-id agent-id status]
+  [stream workflow-id agent-id status & [opts]]
   (-> (create-envelope stream :control-plane/agent-heartbeat workflow-id
                        (str "Heartbeat from " agent-id ": " (name status)))
       (assoc :cp/agent-id agent-id
-             :cp/status status)))
+             :cp/status status)
+      (cond->
+        (:task opts) (assoc :cp/task (:task opts))
+        (:metrics opts) (assoc :cp/metrics (:metrics opts)))))
 
 (defn cp-agent-state-changed
   "Emit when an agent's normalized state changes."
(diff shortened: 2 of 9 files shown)


components/adapter-miniforge/test/ai/miniforge/adapter_miniforge/interface_test.clj
```diff
@@ -77,7 +77,8 @@
         (is (= #{:code-generation :test-writing :code-review}
                (:agent/capabilities info)))
         (is (= {:workflow-id "wf-123"
-                :event-type  :agent-started}
+                :event-type  :agent-started
+                :message "Starting agent"}
```
H31MDALLR, line 81: This should be localized .. this message. It may not always be English.

components/event-stream/src/ai/miniforge/event_stream/core.clj
```diff
@@ -594,21 +603,30 @@
...
+               priority-or-opts
+               {:priority priority-or-opts})]
   (-> (create-envelope stream :control-plane/decision-created workflow-id
                        (str "Decision needed from " agent-id ": " summary))
       (assoc :cp/agent-id agent-id
              :cp/decision-id decision-id
              :cp/summary summary)
-      (cond-> priority (assoc :cp/priority priority))))
+      (cond->
+        (:priority opts) (assoc :cp/priority (:priority opts))
```
H31MDALLR, line 616: These patterns can be better served by extracting the values in a let block and then reusing them in all of the rest of these cond expressions. Or create predicates that test for these things.

PR #700, opened 2026-04-30
test(dag-executor): cover parallel.clj — fix infinite-batch bug + close requiring-resolve race

H31MDALLR:
## Summary

Continues the \`dag-executor\` coverage pass (#688, #695). Adds 34 tests (92 assertions) for \`parallel.clj\` — the lock pool, file-lock conflict detection, parallel batch selection, capacity reporting, and the \`release-all-locks!\` cleanup path.

**Two production bugs caught and fixed by writing the tests.**

## Bug 1: \`select-parallel-batch\` returns nils when ready < max-parallel

The loop used \`(nil? remaining)\` to terminate, but \`(rest seq)\` returns an empty seq \`()\`, not nil. With fewer ready tasks than \`max-parallel\`, the loop kept iterating: \`(first '())\` → nil → falls through the conflict-check (nil's empty file-set never conflicts) → \`conj\`s nil into \`selected\` until \`(>= count max-parallel)\` finally fired.

So a batch of 3 ready tasks asked to run with \`max-parallel 10\` returned:
\`\`\`
[a c nil nil nil nil nil nil nil nil]
\`\`\`
Three real tasks plus seven nils. The downstream scheduler would have either crashed on a nil task-id or silently no-op'd.

**Fix:** \`(empty? remaining)\` covers both the nil and the empty-seq case, terminating cleanly when the input is exhausted.

## Bug 2: \`emit-task-state-event!\` \`requiring-resolve\` race

Same pattern that #692 closed in \`web-dashboard\` handlers. Adding any new test file to the \`dag-executor\` brick increased classloader pressure enough to make the \`requiring-resolve\` calls in \`emit-task-state-event!\` intermittently return nil — silently skipping event emission and breaking \`transition-task!-emits-event-test\` roughly 50% of \`bb pre-commit\` runs.

**Fix:** Add \`event-stream\` as a hard dep in \`dag-executor/deps.edn\` and replace the \`requiring-resolve\` plumbing with a direct \`require\`. Same shape as #692. Verified: **8/8 clean \`bb test\` runs** after the fix.

### Layer 1 — primitives
- \`resource-types\` exposes the documented closed set
- \`create-lock-pool\` defaults (1 repo-write, 4 worktrees) + custom overrides
- \`create-lock\` field shape; \`:lock/files\` set only for \`:exclusive-files\`
- \`files-overlap?\` across sets, vectors, and nil edges

### Layer 1 — repo-write
- Success path: lock recorded under holder-id, capacity drops by 1
- Contention timeout: with \`max-repo-writes 1\`, second holder times out at \`0ms\`
- Release returns the permit and clears the holder entry
- \`:lock-not-found\` on spurious release
- \`:lock-not-found\` when holder owns a different lock type

### Layer 1 — file locks
- Success: files claimed for the holder
- Overlap conflict: \`:lock-conflict\` reports the holding agent in \`:conflicts\`
- Non-overlapping holders coexist
- Same-holder re-acquire replaces (current semantics, documented)
- Release clears both \`:locks\` and \`:file-locks\`
- \`:lock-not-found\` on spurious release

### Layer 1 — worktree
- Success, capacity timeout, permit return, spurious-release error

### Layer 2 — \`can-run-parallel?\`
- No-locks default ⇒ true
- Overlapping file-locks ⇒ false
- Both \`:needs-repo-write?\` ⇒ false
- One repo-write + one non-write ⇒ true

### Layer 2 — \`select-parallel-batch\`
- \`max-parallel\` honored regardless of ready count
- Overlapping \`:exclusive-files\` prevent co-selection
- Empty ready set ⇒ \`[]\`
- \`max-parallel 0\` ⇒ \`[]\` even with ready tasks

### Layer 2 — \`available-capacity\`
- Tracks all three counters (repo-write, worktree, file-locks) across acquisition

### Layer 2 — \`release-all-locks!\`
- Clears repo-write + capacity returned
- Clears worktree + capacity returned
- No-op for non-holder
- File-locks-only path (no semaphore involved)

## Context

PR 8 of the multi-PR test-coverage and standards-drift pass. Builds on #678, #679, #684, #685, #686, #688, #695.

## Test plan

- [x] \`bb pre-commit\` passes — lint + format clean, 2908 tests / 10790 passes / 0 failures / 0 errors, GraalVM clean
- [x] All 34 new tests pass directly via the \`:test\` alias
- [x] **Verified deterministic across 8/8 \`bb test\` runs** after the requiring-resolve fix landed (was ~50% flake before)

### DIFF
--- a/components/dag-executor/deps.edn
+++ b/components/dag-executor/deps.edn
@@ -6,7 +6,8 @@
         ai.miniforge/logging {:local/root "../logging"}
         ai.miniforge/schema {:local/root "../schema"}
         ai.miniforge/agent {:local/root "../agent"}
-        ai.miniforge/artifact {:local/root "../artifact"}}
+        ai.miniforge/artifact {:local/root "../artifact"}
+        ai.miniforge/event-stream {:local/root "../event-stream"}}
  :aliases {:test {:extra-paths ["test"]
                   :extra-deps {io.github.cognitect-labs/test-runner
                                {:git/tag "v0.5.1" :git/sha "dfb30dd"}}}}}
--- a/components/dag-executor/src/ai/miniforge/dag_executor/parallel.clj
+++ b/components/dag-executor/src/ai/miniforge/dag_executor/parallel.clj
@@ -261,7 +261,10 @@
            selected []
            selected-files #{}]
       (cond
-        (nil? remaining)
+        ;; (rest seq) returns an empty seq (), not nil — `empty?` covers both
+        ;; the initial nil case (no ready tasks) and the post-iteration empty
+        ;; case (otherwise we'd append nils until max-parallel fires).
+        (empty? remaining)
         selected
 
         (>= (count selected) max-parallel)
--- a/components/dag-executor/src/ai/miniforge/dag_executor/state.clj
+++ b/components/dag-executor/src/ai/miniforge/dag_executor/state.clj
@@ -26,6 +26,7 @@
   (:require
    [ai.miniforge.dag-executor.result :as result]
    [ai.miniforge.dag-executor.state-profile :as profiles]
+   [ai.miniforge.event-stream.interface :as es]
    [ai.miniforge.logging.interface :as log]
    [clojure.set :as set]))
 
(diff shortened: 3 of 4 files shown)


H31MDALLR:
@copilot resolve the merge conflicts in this pull request

PR #760, opened 2026-05-03
fix: harden claude dogfood repair loop and observability

H31MDALLR:
## Summary

This PR packages the next set of Claude dogfood fixes from the `behavioral-verification-monitor` run on `main`.

The run surfaced three real product issues:

1. repair attempts could claim `:already-implemented` without artifact evidence and slip through phase boundaries
2. active workflows could be marked stale while agent and tool activity was still flowing
3. CLI status output ignored `MINIFORGE_HOME`, so isolated dogfood runs printed the wrong events directory

It also exposed two supporting runtime issues that made diagnosis slower than it should have been:

1. tool-hook stdin handling could deadlock on unbounded reads
2. planner / LLM streaming fallback behavior could look like a silent hang even when the backend had already failed or
  returned empty streaming output

### Repair loops now fail closed

- tighten implementer parsing so prose-only or bare `:already-implemented` claims are not accepted as successful repair
  output
- prefer real file artifacts over status claims when both are present
- reject `:already-implemented` during repair attempts unless there is concrete artifact evidence
- add an `implement` phase boundary check so prior review/verify failures cannot be bypassed by an unverified
  `:already-implemented` handoff

### Planner and runtime observability are more reliable

- stop tool-hook stdin from blocking on full-stream `slurp`
- accept planner-submitted plan artifacts even when the backend marks the response unsuccessful
- recover EDN plan submissions from failed stdout when the planner clearly produced a plan
- add a single retry path for “analysis completed but no plan submission” outcomes
- treat blank successful streaming responses as a fallback case instead of a long fake stall
- preserve stderr diagnostics and shorten the post-kill wait path for stalled planner subprocesses
- print tool activity in the terminal callback path so tool-heavy phases do not look dead

### Workflow freshness now tracks real activity

- update supervisory-state workflow freshness on agent, tool, gate, LLM, and task activity events
- stop deriving false `Workflow stale` attention while a long-running phase is actively doing work

### CLI path reporting now honors isolated homes

- make CLI app-config prefer `MINIFORGE_HOME` for its home/events/config paths
- keep the profile-based path as the fallback when the env override is unset

## Validation

- focused agent / implement / planner / LLM / event-stream tests passed during development
- focused lint for the new CLI and supervisory-state fixes passed
- focused workspace tests passed:
  - `ai.miniforge.cli.app-config-test`
  - `ai.miniforge.supervisory-state.accumulator-test`
- live Claude dogfood rerun confirmed:
  - planner uses MCP context tools during exploration
  - bogus repair-loop `:already-implemented` drift no longer recurs
  - review redirects back to implement with concrete findings
  - nested tool activity is visible again instead of looking fully silent

## Follow-up

This PR should be followed by another Claude dogfood rerun on `main` to confirm the stale-attention and home-dir fixes
in the same end-to-end workflow run.


components/agent/src/ai/miniforge/agent/implementer.clj
```diff
@@ -387,6 +384,22 @@
...
+(defn- repair-attempt?
+  "Return true when the implementer is handling prior review or verify feedback."
+  [input]
+  (or (:task/review-feedback input)
+      (:task/verify-failures input)))
+
+(defn- unverified-already-implemented-response
+  "Reject an :already-implemented claim when there is no artifact evidence on a repair attempt."
+  [input tokens]
+  (response/error
```
H31MDALLR, line 396: Make sure all strings are localized.

components/agent/src/ai/miniforge/agent/planner.clj
```diff
@@ -165,6 +173,41 @@
...
+  "Return the best available content payload for plan extraction."
+  [llm-response]
+  (or (llm/get-content llm-response)
+      (get (llm/get-error llm-response) :stdout)
+      ""))
+
+(defn- submission-retry-prompt
+  "Build a short, submission-only retry prompt for planner recovery."
+  [spec-text prior-content]
+  (str "You already completed the planning analysis for this specification.\n\n"
```
H31MDALLR: This should be a template (data) and not code.

components/agent/src/ai/miniforge/agent/planner.clj
```diff
@@ -365,14 +424,80 @@
...
+
+      (:num-turns llm-response)
+      (assoc :num-turns (:num-turns llm-response)))))
+
+(defn- invoke-planner-submission-retry-session
+  "Run one short follow-up turn that only submits the final plan."
+  [session llm-client retry-prompt config context on-chunk]
+  (let [budget-usd (budget/resolve-cost-budget-usd :planner config context)
+        retry-monitor (llm/create-progress-monitor
+                       {:stagnation-threshold-ms 30000
```
H31MDALLR: No magic numbers or hardcoded configuration (which should be data)

H31MDALLR:
## Review-comment + principle-gap pass

Two new commits address every line-comment and audit gaps against the project principles.

### `ee8be741` — review comments + first-pass gaps
- **H31MDALLR / planner.clj:186 (template-as-data)** — submission-retry prompt moved to `planner.edn` `:prompt/submission-retry-template` with `{{spec-text}}` / `{{prior-content}}` markers.
- **H31MDALLR / planner.clj:469 (no magic numbers)** — main-turn and retry-turn progress monitors plus the retry max-turns moved to `planner.edn` (`:prompt/progress-monitor`, `:prompt/submission-retry-monitor`, `:prompt/submission-retry-max-turns`).
- **H31MDALLR / implementer.clj:396 (localize)** — hardcoded English error replaced with `(messages/t :error/unverified-already-implemented)`; same fix applied to `phase-software-factory/implement.clj`'s `build-unverified-already-implemented-error`.
- **Copilot / user.clj + app_config.clj** — `find-resource`, `getenv`, `default-home-dir` made public (with docstrings flagging them as test seams). Tests now compile cleanly across Clojure implementations without depending on `with-redefs`'s implicit `(var sym)` expansion.
- **Copilot / llm_client.clj:575 (future leak)** — `read-line-with-timeout` now closes the reader and cancels the future on timeout (so the blocked `.readLine` thread exits); removed the dead `catch TimeoutException` (deref-with-default doesn't throw).
- **Copilot / planner.clj:588 (stale telemetry)** — `recover-submitted-plan` now returns the post-retry `:plan-artifact`; `create-planner` invoke-fn threads `final-plan-artifact` into `planner-log-data` so the log reflects the retry outcome rather than the initial response.

**Additional gap fix in the same commit:** the planner user-turn prompt was a 23-line inline `(str ...)` with multiple hardcoded English blocks. Moved to `planner.edn` `:prompt/user-template` + `:prompt/existing-files-template`. Added a small `render-template` helper using `{{key}}` markers; `submission-retry-prompt` now uses it too for symmetry.

### `576ce158` — second-pass principle gaps
- **`llm_client.clj` magic numbers** — extracted `default-stagnation-threshold-ms`, `default-max-total-ms`, `default-min-activity-interval-ms`, `stream-line-timeout-ms`, `process-join-timeout-ms`, `post-kill-join-timeout-ms` as named consts whose docstrings carry the observation history that justified each value.
- **`event-stream/callbacks.clj` localization** — the streaming `[tool] <names>` terminal preview line and the legacy `"Agent calling tool"` status payload (the last raw English strings on this path) now go through the event-stream catalog as `:stream/tool-use-line` / `:stream/agent-calling-tool`.
- **`agent.model` `requiring-resolve` race** — `resolve-llm-client-for-role` used three `requiring-resolve` calls inside a let; under `bb test`'s pmap parallelism the resolve raced and intermittently returned nil (~1 of 3 runs throws NPE). Per `feedback_avoid_requiring_resolve`: replaced with a regular `:require [ai.miniforge.llm.interface :as llm]` and added `ai.miniforge/llm` to `agent/deps.edn`. Verified four consecutive `bb test` runs clean.

### Verification
- `bb test`: 4793 tests, 22248 passes, 0 failures, 0 errors (4× consecutive after `model.clj` de-flake).
- All 9 review threads marked resolved.

PR #761, opened 2026-05-03
fix(implement): don't flag :already-implemented as degraded handoff

H31MDALLR:
## Summary

Surfaced during the codex/event-log-tool-visibility dogfood (2026-05-02). Complements PR #760's stricter `:already-implemented` rejection in `leave-implement`; this fixes a different layer.

## What changed

`phase-software-factory/implement.clj` — narrow when the result-cond flags `:degraded-handoff?` from "anything that isn't `:status :success`" to specifically `:status :error`.

## Why

`phase/result-succeeded?` returns true only for `:status :success`, so `:status :already-implemented` (a deliberate, correct skip when prior-turn work is intact) was being lumped in with real failures. With the curator successfully recovering files from disk, the result was force-promoted to `:status :success` BUT also tagged `:code/degraded-handoff? true`. The reviewer's `implementation-handoff-gate` then emitted a blocking error for any artifact carrying that tag, overriding the LLM-reviewer's `:approved` verdict via `merge-gate-overrides` → final decision `:rejected` regardless of LLM verdict → `:review-approved` gate fails → DAG cascades.

`:already-implemented` is a success state semantically (the agent reported, correctly, that the work was already done) and must not be conflated with a failed handoff.

## Composes with PR #760

- **First iteration** — agent says `:already-implemented` → this fix prevents a false `:degraded-handoff?` tag → success path proceeds.
- **Repair iteration** — agent says `:already-implemented` → this fix still prevents a false tag, but PR #760's `unverified-already-implemented?` check in `leave-implement` correctly fails the phase when there's no artifact evidence of the claimed prior work. Both are needed.

## Test plan

- [x] `bb test` — 4893 tests, 22552 passes, 0 failures, 0 errors.
- [x] Cherry-picked clean onto current `main` (one-line gate change).
- [x] PR #760's regression test (`implement-test/reject-already-implemented-after-prior-review-feedback-test`) still passes — its `agent/invoke` mock returns `response/error`, which my fix's `impl-errored?` check still recognizes.

### DIFF
--- a/components/agent/resources/prompts/planner.edn
+++ b/components/agent/resources/prompts/planner.edn
@@ -5,7 +5,7 @@
 ;; creates detailed implementation plans.
 
 {:prompt/id :planner
- :prompt/version "2.3.0"
+ :prompt/version "2.4.0"
  :prompt/agent :planner
 
  ;; Hard turn cap passed to the backend CLI. Observed 2026-04-18: at 40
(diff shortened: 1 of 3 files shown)


H31MDALLR:
## Copilot review pass — `fbbbc3a0`

All three Copilot threads addressed:

1. **`impl-errored?` only checked `:status :error`** — now uses `response/error?`, which recognises every error shape the implementer can return: `:status :error`, `:status :failed` (legacy phase shape), and `:success false` (output of `response/failure`, e.g. when `agent/invoke` throws). The original strict check would have silently let exception-shaped handoffs slip past the degraded flag.

2. **Removed the now-unused `impl-succeeded?` let-binding** — dead since the cond switched to `impl-errored?`.

3. **Added two regression tests:**
   - `implement-marks-curator-recovered-failure-shape-as-degraded-handoff-test` — guards (1) so `response/failure` / `:success false` always tags degraded, and a future tightening of `response/error?` to `:status`-only doesn't silently regress.
   - `implement-does-not-mark-already-implemented-as-degraded-handoff-test` — guards the whole reason for this PR: a curator-recovered `:already-implemented` outcome must not be flagged degraded, otherwise the reviewer's handoff gate auto-rejects every legitimate "already done" repair.

`bb test`: 4895 tests, 22555 passes, 0 failures, 0 errors.

PR #762, opened 2026-05-03
feat(review): heuristic 'useless loop' detector for repair iterations

H31MDALLR:
## Summary

Surfaced during the codex/event-log-tool-visibility dogfood: with the prior gating fixes (PRs #760 / #761) the workflow ran **three full review-implement cycles** before exhausting the iteration budget — but in two of those three cycles the reviewer's blocking complaints were essentially the same issues, just on slightly different code.

Bumping the iteration cap would only buy more burn. **Detecting stagnation is the actual signal.**

## Approach

Heuristic: reduce each review to a stable fingerprint of its actionable issues, compare against the prior iteration's fingerprint, and short-circuit the loop when they match.

### `15ec32b2` — pure helpers in `agent/reviewer.clj`
- **`review-fingerprint`** — sorted vec of `[severity file line description-hash]` tuples for `:blocking` + `:warning` issues only. `:nit` is intentionally excluded (long-tail polish, not progress signal).
- **`stagnated?`** — strict equality between prior and current fingerprints. v1 starts strict; loosening to set-overlap with a threshold belongs in a follow-up once we have data on real-world false-positive rates.
- Re-exported through `agent.interface` as `review-fingerprint` / `review-stagnated?` so callers don't reach past the polylith interface.

### `8aa19121` — `leave-review` wiring in `phase-software-factory`
- Tracks every review's fingerprint on `[:execution :review-fingerprints]` (vec, oldest-first; lives under `:execution` so it survives the per-phase ctx-clear).
- When the reviewer is blocked AND the new fingerprint matches the immediately prior one, terminates with `:anomalies.review/stagnation` carrying the fingerprint chain — instead of redirecting back to `:implement` to burn another cycle.
- Helper extraction (`terminate-stagnated`, `redirect-to-implement`, `record-fingerprint`) so `leave-review` reads as a pipeline of named decisions rather than the prior nested if/cond.
- Localized `:review/stagnation` message.

## Behavior preserved for

- **First iteration** (no prior fingerprint) — no short-circuit; first review never stagnates.
- **Approved review with non-blocking issues** — empty fingerprint, not stagnation.
- **Real progress** (one issue resolved between iterations) — different fingerprint, ordinary repair redirect.
- **Iteration-budget exhaustion** — unchanged.

## Test plan

- [x] `bb test` — 4908 tests, 22579 passes, 0 failures, 0 errors (13 new across both slices).
- [x] `review-fingerprint` covers actionable severities, excludes nits, is order-independent, distinguishes description changes, and returns `[]` when nothing actionable.
- [x] `stagnated?` returns true on identical non-empty fingerprints, false on first iteration, false on empty input, false on real progress.
- [x] `leave-review` integration: stagnation terminates with the anomaly + fingerprint chain attached, progress still redirects, first iteration never short-circuits, fingerprint history is appended on every iteration.
- [ ] Dogfood validation — run on `event-log-tool-visibility.spec.edn` once this lands and confirm the loop bails on stagnation rather than burning the iteration budget.

### DIFF
--- a/components/phase-software-factory/src/ai/miniforge/phase_software_factory/review.clj
+++ b/components/phase-software-factory/src/ai/miniforge/phase_software_factory/review.clj
@@ -23,8 +23,9 @@
    Agent: :reviewer
    Default gates: [:review-approved :quality-check]"
   (:require            [ai.miniforge.phase.interface :as phase]
+            [ai.miniforge.phase-software-factory.messages :as messages]
             [ai.miniforge.phase-software-factory.phase-config :as phase-config]
-            
+
             [ai.miniforge.phase-software-factory.knowledge-helpers :as kb-helpers]
             [ai.miniforge.agent.interface :as agent]
             [ai.miniforge.knowledge.interface :as knowledge]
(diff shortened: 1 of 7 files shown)


components/phase-software-factory/src/ai/miniforge/phase_software_factory/review.clj
```diff
@@ -149,60 +150,103 @@
...
-               (:knowledge-store ctx) :reviewer
-               (get-in ctx [:execution/input :title]) feedback)
-              (let [phase-result (-> (:phase updated-ctx)
-                                     (update :iterations (fnil inc 1))
-                                     (assoc :review-feedback feedback)
-                                     (phase/request-redirect :implement))]
-                (assoc updated-ctx :phase phase-result)))
+        prior-history     (get-in ctx [:execution :review-fingerprints] [])
+        current-fp        (agent/review-fingerprint review-artifact)
+        stagnated?        (and reviewer-blocked?
```
H31MDALLR: Pull all the left forms that wrap a line and are bloating this function out into their own functions at the right strata.

components/phase-software-factory/src/ai/miniforge/phase_software_factory/review.clj
```diff
@@ -149,60 +150,103 @@
...
+                              (assoc-in [:metrics :review :repair-cycles] (dec iterations))
+                              ;; Merge agent metrics into execution metrics
+                              (update-in [:execution/metrics :tokens] (fnil + 0) (:tokens metrics 0))
+                              (update-in [:execution/metrics :duration-ms] (fnil + 0) (:duration-ms metrics 0))
+                              (record-fingerprint current-fp))]
+    (when redirect?
+      (knowledge/capture-feedback-learning!
+       (:knowledge-store ctx) :reviewer
+       (get-in ctx [:execution/input :title]) feedback))
+    (doto (cond
```
H31MDALLR: Why are we using doto here? I don't see any Java objects in play.

components/phase-software-factory/src/ai/miniforge/phase_software_factory/review.clj
```diff
@@ -149,60 +150,103 @@
...
+            stagnated?
+            (terminate-stagnated updated-ctx
+                                 (conj prior-history current-fp))
+
+            redirect?
+            (redirect-to-implement updated-ctx feedback)
+
+            :else
             (cond-> updated-ctx
```
H31MDALLR: It seems like an `if`-statement or a `when`.

components/phase-software-factory/test/ai/miniforge/phase_software_factory/review_repair_loop_test.clj
```diff
@@ -172,3 +172,92 @@
...
+  {:severity :blocking :file "src/bar.clj" :line 4 :description "Worse"})
+
+(defn- run-leave-review
+  "Run leave-review against a custom ctx and return the resulting full ctx."
+  [{:keys [issues iterations max-iterations prior-fingerprints]
+    :or   {iterations 1 max-iterations 4 prior-fingerprints []}}]
+  (let [result {:output  {:review/decision :changes-requested
+                          :review/issues issues}
+                :metrics {:tokens 100 :duration-ms 5000}}
+        ctx {:phase {:started-at (- (System/currentTimeMillis) 1000)
```
H31MDALLR: No Magic numbers.

components/phase-software-factory/test/ai/miniforge/phase_software_factory/review_repair_loop_test.clj
```diff
@@ -172,3 +172,92 @@
...
+                     :result result}
+             :phase-config {:phase :review}
+             :execution {:review-fingerprints prior-fingerprints}
+             :execution/phase-results {}
+             :execution/input {:description "test task"}
+             :execution/metrics {}}
+        interceptor (phase/get-phase-interceptor {:phase :review})]
+    ((:leave interceptor) ctx)))
+
+(deftest stagnation-terminates-instead-of-redirecting-test
```
H31MDALLR, line 229: Remove all the magic numbers from this PR, whether they're in test or in code, because they're both code.

H31MDALLR:
## Review-comment + magic-number pass — `d933601f` + `3b4c6700`

All eight threads addressed across two commits, sliced under the 200-line commit budget.

### `d933601f` — `agent.reviewer` fingerprint hardening (Copilot 2 + 3)

- **`issue-fingerprint` no longer hashes the description.** Used `(hash description)` to keep tuples compact, but Clojure's `hash` (and the underlying `String#hashCode`) is not collision-free — a theoretical collision would surface as a *false-positive stagnation termination*. Switched to the whitespace-normalized full description string. Descriptions are small enough that the memory cost is negligible and equality becomes exact.
- **`review-fingerprint` now also fingerprints failed gates.** In gate-only mode the reviewer populates `:review/gate-results` without filling `:review/issues`, so the fingerprint would be empty even when blocked → stagnation guard would never fire on a looping gate failure. Added `failed-gate-fingerprint` that converts each failed gate into a virtual `[:blocking ":gate/<gate-id>" 0 description]` tuple.
- New tests: `review-fingerprint-survives-whitespace-reformatting-test`, `review-fingerprint-includes-failed-gates-test`.

### `3b4c6700` — `phase-software-factory` orchestration cleanup (H31MDALLR + Copilot 1)

- **Pull the long inline forms out** (H — review.clj:215). `compute-stagnated?`, `compute-phase-status`, `compute-decision`, `apply-decision`, `accumulate-base-ctx`, `mark-completed`, `review-feedback`. `leave-review` now reads as a pipeline: compute decision data → apply always-on ctx updates → dispatch on decision → emit telemetry → return.
- **Drop the `doto`** (H — review.clj:241). It was being used to thread `next-ctx` into `emit-phase-completed!` (which needs the ctx for the event-stream lookup), but `doto` reads as imperative Java side-effects. Replaced with explicit threading: `(emit-phase-completed! next-ctx :review {...})` and return `next-ctx`.
- **Drop the single-branch `cond->`** (H — review.clj:250). Only the `:completed` case ever fired; replaced with a dedicated `mark-completed` helper, dispatched via `apply-decision`'s `:complete` branch.
- **`terminate-stagnated` now sets `:message` alongside `:anomaly/message`** (Copilot 1). Display/diagnostic consumers read the bare `:message` key in several places — surfacing the stagnation as "Unknown" would defeat the anomaly.
- **`blocking-decisions` named def** replaces the inline `#{:rejected :changes-requested}` set so the intent (and the iter-23 regression reason) reads as one anchored thing.
- **Magic numbers in tests removed** (H — `:196` and `:208`). Every `1`, `2`, `4`, `100`, `1000`, `5000` in the new stagnation tests now has a named constant: `first-iteration`, `second-iteration`, `default-max-iterations`, `mock-token-count`, `mock-elapsed-ms`, `mock-duration-ms`, `one-fingerprint`, `two-fingerprints`. Also dropped the `(hash "seed")` literal in `seed-fp` since the fingerprint shape no longer hashes.
- Added an assertion that the stagnation `:phase :error` carries `:message` (covers Copilot 1).

### Verification

- `bb test`: 4910 tests, 22584 passes, 0 failures, 0 errors.
- All 8 review threads resolved.

PR #763, opened 2026-05-04
diag(release): surface why the release phase fails

H31MDALLR:
## Summary

The 2026-05-03 codex dogfood on `event-log-tool-visibility.spec.edn` walked the full SDLC for the first time (plan → implement → verify → review **`:approved`** → release) but the release phase failed in **0ms** with zero actionable detail:

- `phase-completed` event carried only `:phase/outcome :failure`.
- Persisted snapshot had only the generic `:anomalies.phase/agent-failed`.
- Dogfood log had **no release-executor lines at all**.

Three visibility holes plugged so the next run gives us the actual cause.

### `components/phase-software-factory/src/ai/miniforge/phase_software_factory/release.clj`

1. **`(catch Exception e (response/failure e))` at `enter-release`** was swallowing the exception silently. Added `log/error :release/executor-threw` carrying `(.getName (class e))`, `(ex-message e)`, and `(ex-data e)` so the next dogfood yields the actual cause instead of the wrapped anomaly.

2. **The `:success? false` branch** built a `response/failure` from `(:errors exec-result)` / `(:metrics exec-result)` but never logged them. Added `log/error :release/executor-failed` before the `response/failure` so the executor's reason hits the event stream.

3. **`build-executor-context` read `:logger` from ctx**, so when the workflow ctx lacked `:execution/logger` the release executor ran with `logger=nil` and emitted zero log lines — exactly what we saw. `phase-started` fired (it's emitted by the phase telemetry layer, not the executor) but the entire pipeline-internal log chain was silent. Plumbed the phase-local `logger` through as a third arg to `build-executor-context` so the executor's `phase-started` / `fail` / `phase-completed` logs are emitted regardless of ctx wiring.

Also added a `log/info :release/executor-invoked` at the call site stamping `worktree-path`, `executor`, `env-id`, `create-pr?`, `releaser-agent?` presence, and artifact count — a sanity-check breadcrumb right before the executor runs.

### Tests

- `release-fails-when-write-fails-test` extended to assert that the executor's `:errors` flow into `[:phase :result :error :data :errors]` — guards against a future refactor that drops them silently.
- New `release-surfaces-thrown-exception-in-result-test` asserts that when `execute-release-phase` throws, both message and `ex-data` are preserved on the wrapped failure so post-mortems work.

## Test plan

- [x] `bb test` — 4911 tests, 22587 passes, 0 failures, 0 errors (added 1; 1 pre-existing flake on `test-run!-returns-result-on-zero-exit` unrelated to this change).
- [ ] Dogfood `work/event-log-tool-visibility.spec.edn` once this lands; the release-phase 0ms fail should now carry an actionable cause in the event log + dogfood stdout.

## Why pure diagnostics

This PR adds no behavioral change beyond observability. The release phase still fails the same way it did on 2026-05-03 — but next run we'll know **why**, which is the prerequisite for the actual fix.

### DIFF
--- a/components/phase-software-factory/src/ai/miniforge/phase_software_factory/release.clj
+++ b/components/phase-software-factory/src/ai/miniforge/phase_software_factory/release.clj
@@ -226,8 +226,14 @@
 (defn build-executor-context
   "Build context for the release executor from phase context.
    Includes :github-token so the release executor can inject GH_TOKEN
-   into the capsule's environment for gh CLI authentication."
-  [ctx config]
+   into the capsule's environment for gh CLI authentication.
+
+   The phase-local `logger` is threaded in so the release executor's
+   log calls (phase-started / fail / phase-completed) are emitted
+   even when the workflow ctx lacks :execution/logger — observed in
+   the 2026-05-03 dogfood, where release failed in 0ms with zero
+   visible log lines because the executor's logger was nil."
+  [ctx config logger]
   (let [on-chunk (phase/create-streaming-callback ctx :release)]
     (cond-> {:worktree-path (or (get-in ctx [:execution/worktree-path])
                                 (get-in ctx [:worktree-path])
@@ -236,7 +242,7 @@
                                 (System/getProperty "user.dir"))
              :executor (get-in ctx [:execution/executor])
              :environment-id (get-in ctx [:execution/environment-id])
-             :logger (get-in ctx [:execution/logger])
+             :logger (or (get-in ctx [:execution/logger]) logger)
              :llm-backend (get-in ctx [:execution/llm-backend])
              :artifact-store (get-in ctx [:execution/artifact-store])
              :event-stream (:event-stream ctx)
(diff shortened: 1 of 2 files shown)


H31MDALLR:
## Copilot was right — `aec220b4`

Both comments were valid: the original PR's assertions only verified the wrapped `:phase :result`, so a refactor that quietly dropped either the log emission or the logger fallback would still have passed the suite while losing the entire observability goal of this PR.

Three new tests pin the production fixes:

1. **`release-passes-non-nil-logger-to-executor-when-ctx-logger-absent-test`** — captures the `exec-context` handed to `execute-release-phase` when the workflow ctx has no `:execution/logger` and asserts `:logger` is non-nil. Pre-fix the executor ran with `logger=nil` whenever the ctx wiring left `:execution/logger` absent — the exact silent-fail path observed in the 2026-05-03 dogfood.

2. **`release-logs-executor-failure-test`** — installs a capturing logger and asserts `:release/executor-failed` fires on the `:success? false` branch.

3. **`release-logs-thrown-exception-test`** — mirrors (2) for the catch branch; `:release/executor-threw` must fire when `execute-release-phase` throws.

Two small helpers (`capturing-logger` + `entry-events`) factored out so future tests in this file can pin emission similarly without duplicating the atom plumbing.

`bb test`: 4914 tests, 22591 passes, 0 failures, 0 errors (added 3).

PR #765, opened 2026-05-04
chore(standards): bump submodule + recompile pack for try+ rule

H31MDALLR:
## Summary

Lands the consumer side of [miniforge-ai/miniforge-standards#15](https://github.com/miniforge-ai/miniforge-standards/pull/15) (merged) — the new `languages/clojure-exception-handling` rule (dewey 211).

The rule says: **prefer `try+` / `throw+`** since slingshot's `try+` is a strict superset of `clojure.core/try` (class-name selectors expand to `(instance? <class> %)` and key-value / predicate / `%`-form selectors are added on top). Plain `try` is allowed in three narrow cases (no slingshot dep + class-only catch; nested in a `try+` for try-finally clarity; REPL / `(comment …)` blocks).

## Changes

1. **Bump `.standards` submodule pointer** `164d143 → 100d112` (the merge commit of standards#15).
2. **Re-run `bb standards:pack`** to compile the new rule into `components/phase/resources/packs/miniforge-standards.pack.edn`. Pack now carries 40 rules; new entry id is `:std/clojure-exception-handling`.

After merge the review phase surfaces the rule when catching exceptions in `components/**/src/**/*.clj` — guidance reaches the agents via the same path as every other compiled standard.

## Test plan

- [x] `bb standards:pack` — compiled 40 rules, 0 failures.
- [x] `bb test` — 4910 tests, 22584 passes, 0 failures, 0 errors (one pre-existing flake `test-run!-returns-result-on-zero-exit` cleared on rerun).
- [ ] After merge, dogfood will pick up the new rule on the next review-phase run (the existing dogfood loop on `event-log-tool-visibility.spec.edn` will exercise it).


H31MDALLR:
## Both Copilot comments addressed — `45047a73`

Source-MDC fix landed in [miniforge-ai/miniforge-standards#16](https://github.com/miniforge-ai/miniforge-standards/pull/16) (merged). Consumer side bumps the submodule pointer + re-packs, plus hardens the compiler so future MDC authors using numbered lists don't fall into the same trap.

### Comment 1 — `:agent-behavior` truncated at "4."

Root cause was in the standards-pack compiler, not the source: `condense-bullets` matched `[-*]\s+` only, so my numbered-list (`1. … 5.`) `## Agent behavior` section fell through to `condense-prose`, which truncated mid-sentence at the 500-char target.

Two-part fix:

1. **Source MDC switched to `-` bullets** (standards#16) with the most load-bearing rules first so the compiler's first-3-bullets selection picks up the right content cleanly. Compiled `:agent-behavior` now reads:

   ```
   - Default new catch sites to `try+`; add the slingshot require if absent.
   - If the throw site is `throw+` / `throw-anomaly!`, the catch site MUST be `try+` — never plain `try` digging into `ex-data` for `:anomaly/category`.
   - Reject `try+` whose every catch is class-only AND the component has no other slingshot use — either commit to `try+` across the component or drop the `+`.
   ```

2. **Compiler hardened** (this commit). Extracted `bullet-line-pattern` as a named def (DRY between `condense-bullets` and `condense-to-length`); regex now matches `[-*]` **or** `\d+\.`. Regression test added to `mdc-compiler-test.clj` that asserts a 5-step numbered list compiles to the first three steps preserved in full, with no bare numbered-prefix mid-sentence truncation.

### Comment 2 — title/enforcement message disagreed with body

Source-MDC frontmatter `description` widened from "`try` only when slingshot is not on the classpath" to "plain `try` only in narrow exemptions (no-slingshot-dep leaf catch, leaf inside `try+`, REPL/`(comment …)`)" — matches the body's three documented cases. Catalog entries in `index.mdc` and `CLAUDE.md` updated accordingly.

Compiled `:rule/title` and `:enforcement/:message` now read:

> Clojure exception handling — prefer slingshot `try+` / `throw+`; plain `try` only in narrow exemptions (no-slingshot-dep leaf catch, leaf inside `try+`, REPL/`(comment …)`)

### Verification

- `bb test` — 4910 tests, 22588 passes, 0 failures, 0 errors (added 1 regression deftest with 4 assertions).
- Compiled pack inspected directly: `:agent-behavior` carries the full first-3-bullets value, title/enforcement reads "narrow exemptions."

PR #767, opened 2026-05-04
fix(cli): preflight inherited backend before dogfood

H31MDALLR:
## Summary

This stacked PR follows `#766` and packages the next Claude dogfood fixes from the `behavioral-verification-monitor`
run on `main`.

The remaining startup/runtime problem had two distinct parts:

1. Miniforge was still trusting that the configured backend CLI was healthy, without proving which binary it was about
   to execute or whether that exact binary could answer a trivial request before entering `:explore`
2. the LLM stream reader could sit in one long blocking read, which made an idle-but-open subprocess look like a silent
   hang because progress supervision could not fire until another line arrived or the process exited

On this host that mattered because the failure is version-specific, not “Claude generally broken”:

- `/Users/chris/.local/bin/claude` `2.1.126` answers the non-interactive probe correctly
- `/opt/homebrew/bin/claude` `2.1.89` wedges in `claude -p` mode

So the product needs to stamp the exact inherited binary path/version it will use and fail before workflow execution if
that resolved backend is unhealthy.

### Backend startup now resolves and stamps the real CLI binary

- resolve the backend command from the inherited `PATH` the same way Miniforge will actually launch it
- print backend name, resolved path, and `--version` output before execution starts
- fail closed when the configured CLI command is missing from `PATH`
- fail closed when the version probe itself is unhealthy

### Claude gets a real non-interactive startup probe

- run a bounded Claude CLI preflight before workflow execution:
  - `claude -p "Reply with exactly {\"ok\":true}" --output-format json --max-turns 1`
- run that probe against the resolved binary path, not a freshly spawned shell lookup
- fail closed with probe diagnostics when that exact CLI binary is wedged, exits non-zero, or returns unexpected output

### Other CLI backends still get a bounded health check

- non-Claude CLI backends use the same resolved binary path with a bounded 10-second completion probe
- the failure payload now includes backend name, resolved path, version, and probe response details

### CLI stream supervision is no longer blind during idle stdout

- `process-stream-lines` now polls the stream queue instead of sitting inside one long blocking `readLine`
- progress-monitor timeout checks can now fire while stdout is idle but the process is still alive
- clean EOF is still treated as clean EOF, not misclassified as a timeout
- blank `exit 0` CLI output continues to fail closed as `empty_success_output`

## Validation

- `clj-kondo` on touched CLI and LLM files: clean
- changed-bricks test sweep from the worktree: no failures in the observed run
- focused regression coverage added for:
  - backend stamping and fail-closed preflight anomalies
  - idle stream supervision while waiting for output
  - clean EOF without synthetic timeouts
  - blank successful CLI output

## Live Behavior

- the source-root / MCP anchoring fix is split into `#766`
- this PR makes the remaining backend issue explicit: Miniforge now tells humans and agents which CLI binary it is
  using before the run starts
- if workflow startup resolves to the older wedged Homebrew Claude binary, Miniforge should now fail before `:explore`
  with the stamped path/version and preflight failure payload instead of drifting into a silent planner stall

## Follow-up

- merge `#766` first so this PR stays a narrow stacked follow-up
- rerun Claude dogfood on `main` after both land to confirm:
  - startup provenance is correct
  - the inherited Claude binary is stamped correctly
  - unhealthy backend selection fails loudly before planner execution


bases/cli/src/ai/miniforge/cli/workflow_runner.clj
```diff
@@ -213,6 +215,218 @@
...
+(defn- resolve-cli-command-path
+  [cmd]
+  (cond
+    (str/blank? cmd) nil
+    (str/includes? cmd "/")
+    (when (executable-file? cmd)
+      (str (fs/absolutize cmd)))
+
+    :else
+    (some (fn [entry]
```
H31MDALLR: Extract this fn to a real defn at the appropriate strata.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
```diff
@@ -213,6 +215,218 @@
     (when (:git-dirty? context)
       (println (display/colorize :yellow "   Warning: source checkout has uncommitted changes")))))
 
+(def ^:private backend-preflight-prompt
```
H31MDALLR: Move configuration into data. This should be in an edn file configuring this component.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
```diff
@@ -213,6 +215,218 @@
...
+                         workdir (assoc :dir workdir))
+                       cmd)
+        result (deref process timeout-ms ::timeout)]
+    (if (= ::timeout result)
+      (do
+        (try
+          (when-let [^Process jp (:proc process)]
+            (.destroyForcibly jp))
+          (catch Exception _ nil))
+        {:out ""
```
H31MDALLR, line 327: This shape seems to be used in multiple places. Pull a factory function out so that we're not hand creating the same map over and over again.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
```diff
@@ -213,6 +215,218 @@
...
+      {:out (:out result)
+       :err (:err result)
+       :exit (:exit result)})))
+
+(defn- read-cli-version
+  [cmd-path]
+  (let [{:keys [out err exit timeout-ms]} (run-cli-command [cmd-path "--version"] backend-version-timeout-ms)]
+    (cond
+      timeout-ms
+      {:success false
```
H31MDALLR: Use standard success and anomaly patterns instead of hand creating success maps.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
```diff
@@ -213,6 +215,218 @@
...
+
+(defn- run-claude-backend-preflight
+  [cmd-path workdir]
+  (let [{:keys [out err exit timeout-ms]} (run-cli-command (claude-preflight-command cmd-path)
+                                                           backend-preflight-timeout-ms
+                                                           :workdir workdir)
+        trimmed (some-> out str/trim)]
+    (cond
+      timeout-ms
+      {:success false
```
H31MDALLR: Same here. We're using hand-created success maps. Fix this through our standard success and anomaly patterns.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
```diff
@@ -213,6 +215,218 @@
...
+               :stderr (some-> err str/trim not-empty)}
+       :exit-code exit})))
+
+(defn- run-backend-preflight!
+  [quiet llm-client context]
+  (when-let [{:keys [backend backend-config cmd cmd-path] :as stamp} (backend-stamp llm-client)]
+    (when (:requires-cli? backend-config)
+      (when-not cmd-path
+        (response/throw-anomaly! :anomalies/incorrect
+                                 (str "Configured backend CLI is not on PATH: " cmd)
```
H31MDALLR: All strings are localized. Either place strings that are not going to be seen by a user in a system locale. All other strings go in EN-US.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
```diff
@@ -213,6 +215,218 @@
...
+               :stdout trimmed
+               :stderr (some-> err str/trim not-empty)}
+       :exit-code exit})))
+
+(defn- run-backend-preflight!
+  [quiet llm-client context]
+  (when-let [{:keys [backend backend-config cmd cmd-path] :as stamp} (backend-stamp llm-client)]
+    (when (:requires-cli? backend-config)
+      (when-not cmd-path
+        (response/throw-anomaly! :anomalies/incorrect
```
H31MDALLR: All three of these, when not forms, look almost identical, with the exception of a small set of parameters. Extract them to a function, and then either use polymorphism, or they could just be a function if it's just parameters that are different. 

So there's quite a bit of state in here. We first have a `when let`, and then we have a `when`, and then we have several `when not` patterns. All of those add n that two to the number of states we have to cover if we want to fully test this function. Refactor this into a means that doesn't need all of these conditionals.

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
```diff
@@ -82,10 +83,15 @@
   "How long to wait on subprocess `deref` after we already classified
    a timeout and force-killed the process. The kill makes joining a
    matter of seconds — without this, we paid an extra 10 minutes per
-   timeout (observed Claude dogfood 2026-05-02: 10-min planner cap
+  timeout (observed Claude dogfood 2026-05-02: 10-min planner cap
    becoming an 18-min wall-clock failure)."
   5000)
 
+(def ^:private stream-poll-interval-ms
```
H31MDALLR: Should be configuration, not code.

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
```diff
@@ -605,60 +619,62 @@
...
-                (when dump-writer (.println dump-writer line-or-timeout) (.flush dump-writer))
-                (on-line line-or-timeout)
-                (recur))
-
-            :else
-            nil))))
-    (when dump-writer (.close dump-writer))
+                       (java.io.FileWriter. dump-path true)))
+        line-queue (LinkedBlockingQueue.)
+        reader-future (future
```
H31MDALLR: Pull this out. This should not be an eight-line form into a let variable. Make this a function of its own right so that it doesn't break up the intent of the parent function

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
```diff
@@ -605,60 +619,62 @@
...
+                now (System/currentTimeMillis)]
+            (cond
+              (instance? Exception line-or-signal)
+              (throw line-or-signal)
+
+              (identical? line-or-signal eof-sentinel)
+              nil
+
+              (some? line-or-signal)
+              (do (reset! last-line-at now)
```
H31MDALLR: Extract this from to a function. It's doing too much inline in the cond.

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
```diff
@@ -605,60 +619,62 @@
...
+                            (.put line-queue e))))]
+    (try
+      (loop []
+        (if-let [t (pm/check-timeout monitor)]
+          (reset! timeout-reason t)
+          (let [line-or-signal (.poll line-queue stream-poll-interval-ms TimeUnit/MILLISECONDS)
+                now (System/currentTimeMillis)]
+            (cond
+              (instance? Exception line-or-signal)
+              (throw line-or-signal)
```
H31MDALLR: Why are we not throwing an anomaly? This seems to be a raw throw.

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
```diff
@@ -605,60 +619,62 @@
...
+        reader-future (future
+                        (try
+                          (loop []
+                            (if-some [line (.readLine out-reader)]
+                              (do (.put line-queue line)
+                                  (recur))
+                              (.put line-queue eof-sentinel)))
+                          (catch Exception e
+                            (.put line-queue e))))]
+    (try
```
H31MDALLR: Use slingshot instead of the base try so that we get better exceptions as data fidelity.

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
```diff
@@ -605,60 +619,62 @@
...
+      (loop []
```
H31MDALLR, line 730: This is an infinite loop. Do we mean it to be an infinite loop? If so, then why is this not a thread function?

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
H31MDALLR: Pulled this into named helpers in `workflow_runner.clj` rather than leaving it inline. The preflight/provenance path now uses extracted functions such as `matching-command-path`, `backend-provenance-lines`, `cli-required-backend-stamp`, and `run-backend-probe` so the top-level flow reads as orchestration instead of embedded mechanics.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
H31MDALLR: Moved the backend-preflight settings into component data at `bases/cli/resources/config/cli/workflow-runner.edn`. The runner now reads prompt, args, and timeout policy from config instead of embedding those values in code.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
H31MDALLR: Replaced the repeated hand-built probe result maps with `success-response` and `failure-response` helpers in `workflow_runner.clj`, and those now wrap the standard response/anomaly shapes consistently.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
H31MDALLR: Normalized this path onto the standard response/anomaly conventions. The runner now builds probe outcomes through the shared helper functions instead of hand-assembling ad hoc success maps.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
H31MDALLR: Same cleanup here: hand-built success maps were removed in favor of the shared response factories so the preflight surface stays consistent across the CLI checks.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
H31MDALLR: Localized the user-facing runtime/backend strings in `bases/cli/resources/config/cli/messages/en-US.edn`, and `workflow_runner.clj` now renders provenance/preflight output through those message keys rather than embedding raw strings.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
H31MDALLR: Refactored `run-backend-preflight!` into named transition helpers instead of stacking nested `when` / `when-not` forms. The flow is now split across `ensure-cli-command-path!`, `versioned-backend-stamp`, `verify-backend-probe!`, and `run-backend-probe`, which cuts the conditional state explosion down substantially.

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
H31MDALLR: Moved the stream/Claude defaults into `components/llm/resources/llm/client-defaults.edn`. `llm_client.clj` now reads those values through `client-default` helpers instead of keeping them as hard-coded constants.

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
H31MDALLR: Pulled the inline dump-writer and line bookkeeping work into dedicated helpers. The stream path now uses functions such as `open-stream-dump-writer`, `record-stream-line!`, and `timeout-signal` so the parent loop stays focused on supervision flow.

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
H31MDALLR: Extracted the conditional stream-signal handling into `process-stream-signal`, plus smaller helpers like `anomaly-signal?`, `eof-signal?`, and `stream-poll-signal`. That removed the large inline `cond` from `process-stream-lines`.

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
H31MDALLR: Changed this to throw anomalies as data. The stream path now turns reader failures into response anomalies and propagates them with `throw+` rather than raw exception throws.

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
H31MDALLR: Switched the stream supervision block to `try+` / `throw+` so the failure path keeps the anomaly payload intact instead of collapsing back to base exception handling.

components/llm/src/ai/miniforge/llm/protocols/impl/llm_client.clj
H31MDALLR: Made the infinite reader loop explicit as a reader-thread function. The loop now lives in `read-stream-loop!`, and `start-stream-reader!` owns the future/thread boundary so the intent is visible in the function structure.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
```diff
@@ -203,21 +206,365 @@
...
+(defn- resolve-cli-command-path
+  [cmd]
+  (cond
+    (str/blank? cmd) nil
+    (str/includes? cmd "/")
+    (when (executable-file? cmd)
+      (str (fs/absolutize cmd)))
+
+    :else
+    (some #(matching-command-path % cmd) (path-entries))))
```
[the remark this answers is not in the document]
H31MDALLR: Switched the PATH lookup to a portable resolver. `workflow_runner.clj` now treats direct paths via a separator check and resolves PATH entries through `babashka.fs/which`, which handles Windows executable resolution as well as Unix PATH lookup. That removes the Unix-only `":"`/literal-name assumption from the earlier implementation.

bases/cli/src/ai/miniforge/cli/workflow_runner.clj
```diff
@@ -203,21 +206,365 @@
...
+                                          :workdir workdir))}
+       model (assoc :model model)))))
+
+(defn- run-generic-backend-preflight
+  [llm-client cmd-path workdir]
+  (let [response (llm/complete (create-preflight-client llm-client cmd-path workdir)
+                               {:prompt (backend-preflight-prompt)
+                                :max-turns 1})]
+    (if (and (llm/success? response)
+             (preflight-success? (:content response)))
```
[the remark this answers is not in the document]
H31MDALLR: Added focused coverage for the non-Claude CLI preflight path in `preflight_test.clj`. The new `run-backend-preflight-exercises-generic-cli-success-path-test` drives the Codex-style branch through `llm/create-client` + `llm/complete` and verifies that the canonical `{\"ok\":true}` payload is accepted instead of being rejected at startup.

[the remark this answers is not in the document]
H31MDALLR: Good catch. I changed the generic CLI preflight to bypass llm/complete entirely and run the backend command directly, then decode raw CLI stdout into assistant text before checking the canonical {"ok":true} payload. For streaming backends like Codex, workflow_runner.clj now uses the configured :stream-parser to accumulate message deltas from the JSON event stream, so a healthy Codex CLI no longer fails preflight just because complete-impl would have returned raw event JSON.

PR #772, opened 2026-05-04
feat(workflow): v2 multi-parent merge wiring (Stage 1B)

H31MDALLR:
## Summary

Stage 1B of v2 per-task base chaining (per [I-DAG-MULTI-PARENT-MERGE.md](https://github.com/miniforge-ai/miniforge/blob/main/specs/informative/I-DAG-MULTI-PARENT-MERGE.md)). Lights up the multi-parent code path that Stage 1A's pure-data primitives (PR #770) shipped.

After this PR, multi-parent DAG plans no longer reject at plan-validation time — they run end-to-end via a deterministic git merge. The conflict path surfaces as a typed anomaly (terminal failure for now); Stage 2 will replace that with an automated resolution sub-workflow per spec §6.1.

## What's in this PR

**`merge-parent-branches!`** in the orchestrator implements the spec §3.2 algorithm:
1. Snapshot parent SHAs at merge time (mutable branch names → immutable SHAs).
2. Collapse duplicate tips via the Stage 1A `dag/collapse-duplicate-tips`.
3. Collapse ancestors via `git merge-base --is-ancestor`; preserve plan order on surviving maximal tips.
4. Refuse unrelated histories — surface `:anomalies/dag-multi-parent-unrelated-histories` rather than using `--allow-unrelated-histories` (spec §6.5).
5. Compute `:merge/input-key` via the Stage 1A `dag/compute-merge-input-key`; namespaced ref `refs/miniforge/dag-base/<run-id>/<task-id>/<input-key>`.
6. Run `git merge -s ort` for 2 effective parents, `-s octopus` for 3+, with the spec §3.1 pinned flags (`--no-edit --no-gpg-sign --no-verify --no-ff -m <generated>`) in a clean temp worktree.
7. On success: push the resulting commit to the namespaced ref and clean up. On conflict: enumerate conflicts, surface a typed anomaly, clean up.

**`task-sub-opts`** now calls `merge-parent-branches!` for multi-parent tasks. The merged ref is passed as `:branch`. Merge anomalies are surfaced via `:dag/merge-anomaly` on opts; `run-mini-workflow` short-circuits before running a doomed sub-workflow.

**`execute-plan-as-dag`** no longer rejects non-forest plans. `validate-dag-forest` survives as an informational helper: multi-parent plans log `:dag/multi-parent-detected` for dashboard plan-quality observability and run through the new merge path.

## Test coverage

- **New integration tests** (`merge_parent_branches_integration_test.clj`, 7 tests / 16 assertions) exercise the helper against real git temp repos:
  - 2-parent disjoint-files happy path (asserts merge commit has 2 parents and both files are in the tree)
  - Ancestor collapse → single-parent fast path
  - Duplicate-tip collapse → single-parent fast path
  - Unrelated histories → typed anomaly
  - Conflict on same-file divergence → typed anomaly with conflict paths
  - Empty registry → fallback to default branch
  - Branch unresolvable in repo → typed anomaly
- **Updated unit tests** in `dag_orchestrator_test.clj`:
  - `task-sub-opts-multi-dep` reflects the v2 anomaly surfacing via `:dag/merge-anomaly`
  - `execute-plan-as-dag-rejects-diamond-test` → `execute-plan-as-dag-accepts-diamond-test-v2`; pins that diamond plans run AND log `:dag/multi-parent-detected`

Full suite: **4941 tests / 22684 passes / 0 failures / 0 errors**.

## What's NOT in this PR (deferred to Stage 2)

- Resolution sub-workflow that automates conflict resolution (spec §6.1).
- Merge-resolution curator (`:curator/markers-not-resolved`, `:curator/recurring-conflict`) per spec §6.1.2.
- Policy-overridable agent prompt scaffolding per spec §6.1.3.

For Stage 1B, conflicts are terminal failures with a typed anomaly. Stage 2 makes them recoverable.

## Spec traceability

| Spec section | Implemented in this PR |
| ------------ | ---------------------- |
| §3.1 pinned merge flags | `pinned-merge-flags` |
| §3.2 step 1 SHA snapshot | `snapshot-parent-shas` |
| §3.2 step 4 ancestor collapse | `collapse-ancestors` |
| §3.2 step 5 strategy selection | `run-merge!` (ort vs octopus by parent count) |
| §3.2 step 6 input-key | `dag/compute-merge-input-key` (Stage 1A re-used) |
| §3.2 step 7-9 merge execution | `merge-parent-branches!` orchestration |
| §6.1 conflict anomaly shape | `run-merge!` non-zero-exit branch |
| §6.2 ancestor parent collapse | `collapse-ancestors` |
| §6.3 duplicate parent tips | `dag/collapse-duplicate-tips` (Stage 1A re-used) |
| §6.5 unrelated histories anomaly | `shared-ancestry?` branch |
| §7.2 namespaced ref | `merge-base-ref-name` |

### DIFF
--- a/components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
+++ b/components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
@@ -40,6 +40,7 @@
    [ai.miniforge.logging.interface :as log]
    [ai.miniforge.phase.interface :as phase]
    [ai.miniforge.workflow.dag-resilience :as resilience]
+   [ai.miniforge.workflow.messages :as messages]
    [babashka.fs :as fs]
    [clojure.java.shell :as shell]
    [clojure.set :as set]
@@ -283,6 +284,168 @@
 ;; happy path; conflict / unrelated-histories surface as terminal anomalies
 ;; that Stage 2 will replace with an automated resolution sub-workflow.
 
+;; Constants -----------------------------------------------------------
+
+(def ^:private merge-base-ref-prefix
+  "Namespace under refs/ where multi-parent merge bases live (spec §7.2)."
+  "refs/miniforge/dag-base")
+
+(def ^:private supported-merge-strategies
+  "Strategies `merge-parent-branches!` knows how to execute today.
+   `:git-merge` ships in Stage 1B; `:sequential-merge` is Stage 4
+   (per spec §4 / §10.11). Plans that explicitly request an unsupported
+   strategy get a typed anomaly rather than silently falling through
+   to `:git-merge` (which would misreport the executed strategy in
+   logs and the resolution-sub-workflow payload)."
+  #{:git-merge})
+
+(def ^:private octopus-merge-min-parents
+  "git's `octopus` strategy is required for 3+ parents. The default
+   `ort` strategy handles only 2-head merges. The threshold lives
+   here so the comparison sites are self-describing — no magic 2 / 3."
+  3)
+
+(def ^:private merge-base-default-max-parents
+  "git merge-base without --octopus only handles 2 parents. For 3+
+   we pass --octopus to find the n-way common ancestor. Same magic
+   number as octopus-merge-min-parents but a different code path,
+   named separately so the rationale is explicit at each site."
+  2)
+
+(def ^:private fallback-run-id
+  "Used when context lacks a workflow-id (test scaffolding mostly).
+   The merge ref namespace requires a non-nil run-id segment."
+  "no-run-id")
+
+;; Anomaly factories ---------------------------------------------------
+;; Each factory takes the minimum data needed and produces the canonical
+;; anomaly shape (`:anomaly/category` + `:anomaly/message` + rich data
+;; under `:merge/*` and `:git/*` keys). All messages route through the
+;; workflow message catalog.
+
+(defn- branch-unresolvable-anomaly
+  "Anomaly: a parent's registered branch could not be rev-parsed in
+   the host repo. Usually the registry is out of sync with the repo
+   (test scaffolding without the actual branches, or a registry
+   carry-over after a force-push that rewrote the branch's tip)."
+  [parent git-result]
+  {:anomaly/category :anomalies/dag-multi-parent-branch-unresolvable
+   :anomaly/message  (messages/t :dag.merge/branch-unresolvable)
+   :merge/parent     parent
+   :git/exit-code    (:exit git-result)
+   :git/stderr       (:err git-result)})
+
+(defn- unrelated-histories-anomaly
+  "Anomaly: parents share no common ancestor. v2 refuses to use
+   --allow-unrelated-histories per spec §6.5; this is almost always a
+   plan-quality / repo-state issue (stray git init, cross-repo branch)."
+  [task-id strategy parents]
+  {:anomaly/category :anomalies/dag-multi-parent-unrelated-histories
+   :anomaly/message  (messages/t :dag.merge/unrelated-histories)
+   :task/id          task-id
+   :merge/parents    parents
+   :merge/strategy   strategy})
+
+(defn- strategy-unsupported-anomaly
+  "Anomaly: plan requested a merge strategy this stage doesn't
+   implement. Echoes the requested strategy and the supported set so
+   the operator/dashboard knows what's available right now."
+  [task-id strategy]
+  {:anomaly/category :anomalies/dag-multi-parent-strategy-unsupported
+   :anomaly/message  (messages/t :dag.merge/strategy-unsupported
+                                 {:strategy strategy})
+   :task/id          task-id
+   :merge/strategy   strategy
+   :merge/supported  supported-merge-strategies})
+
+(defn- conflict-anomaly
+  "Anomaly: git merge produced conflicts. Carries the parents,
+   conflict paths, strategy, input-key, and raw git diagnostics.
+   Stage 2 will use this shape as the resolution-sub-workflow input
+   per spec §6.1."
+  [task-id strategy parents conflicts input-key git-result]
+  {:anomaly/category :anomalies/dag-multi-parent-conflict
+   :anomaly/message  (messages/t :dag.merge/conflict)
+   :task/id          task-id
+   :merge/parents    parents
+   :merge/conflicts  conflicts
+   :merge/strategy   strategy
+   :merge/input-key  input-key
+   :git/exit-code    (:exit git-result)
+   :git/stderr       (:err git-result)})
+
+(defn- worktree-setup-failed-anomaly
+  "Anomaly: couldn't create the temp worktree for the merge attempt.
+   Usually a filesystem / git-state issue from a prior crashed run."
+  [task-id git-result]
+  {:anomaly/category :anomalies/dag-multi-parent-merge-failed
+   :anomaly/message  (messages/t :dag.merge/merge-failed-worktree)
+   :task/id          task-id
+   :git/exit-code    (:exit git-result)
+   :git/stderr       (:err git-result)})
+
+(defn- ref-write-failed-anomaly
+  "Anomaly: merge succeeded but `git update-ref` for the namespaced
+   ref failed. Surfacing this matters because returning a 'success'
+   result with an unresolvable :merge/ref would mislead downstream
+   tasks."
+  [task-id ref-name commit-sha git-result]
+  {:anomaly/category :anomalies/dag-multi-parent-merge-failed
+   :anomaly/message  (messages/t :dag.merge/merge-failed-ref-write)
+   :task/id          task-id
+   :merge/ref        ref-name
+   :merge/commit-sha commit-sha
+   :git/exit-code    (:exit git-result)
+   :git/stderr       (:err git-result)})
+
+(defn- merge-error?
+  "True when a value is one of our merge anomalies (the canonical
+   shape with :anomaly/category). Lets callers branch on the result
+   without coupling to specific categories."
+  [x]
+  (and (map? x) (some? (:anomaly/category x))))
+
+;; Result factories ----------------------------------------------------
+;; Successful merge outcomes go through the existing dag/ok result-monad
+;; constructor so consumers can use the same dag/ok? / dag/unwrap
+;; predicates that the rest of the orchestrator already uses for task
+;; results. The inner shape is consistent across the three success
+;; varieties (full merge / single-parent fast-path / fallback) so the
+;; consumer only needs to look for :branch and (optionally) :commit-sha.
+
+(defn- merge-ok-result
+  "Successful merge commit on the namespaced ref. Includes the input-key
+   and a `:cache-hit?` flag for observability."
+  [{:keys [ref-name commit-sha input-key strategy parents collapsed cache-hit?]}]
+  (dag/ok (cond-> {:branch       ref-name
+                   :commit-sha   commit-sha
+                   :input-key    input-key
+                   :strategy     strategy
+                   :parents      parents
+                   :cache-hit?   (boolean cache-hit?)}
+            (seq collapsed) (assoc :collapsed collapsed))))
+
+(defn- single-parent-fast-path-result
+  "Successful resolution where collapse left exactly one effective
+   parent — no merge commit needed; downstream task forks off the
+   surviving parent's branch directly. Spec §6.2 / §6.3."
+  [survivor collapsed]
+  (dag/ok (cond-> {:branch         (:branch survivor)
+                   :commit-sha     (:commit-sha survivor)
+                   :single-parent? true}
+            (seq collapsed) (assoc :collapsed collapsed))))
+
+(defn- empty-registry-fallback-result
+  "Successful (defensive) resolution where the registry has no parents
+   for the declared deps. Mirrors the v1 single-parent path's
+   fail-soft semantics; production scheduling should prevent this."
+  [default-branch]
+  (dag/ok {:branch         default-branch
+           :single-parent? true
+           :fallback-reason :no-registered-parents}))
+
+;; Helpers -------------------------------------------------------------
+
 (defn- run-git
   "Invoke git in `cwd` with `args`. Returns `{:exit :out :err}` matching
    `clojure.java.shell/sh`. Defensive against shell exceptions so callers
@@ -314,11 +477,7 @@
         (if (zero? (:exit r))
           (recur (rest remaining)
                  (conj! out (assoc p :commit-sha (str/trim (:out r)))))
-          {:anomaly/category :anomalies/dag-multi-parent-branch-unresolvable
-           :anomaly/message "Parent branch could not be resolved to a commit"
-           :merge/parent p
-           :git/exit-code (:exit r)
-           :git/stderr (:err r)}))
+          (branch-unresolvable-anomaly p r)))
       {:parents (persistent! out)})))
 
 (defn- ancestor-of?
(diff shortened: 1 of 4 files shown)


components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,295 @@
...
+                              parents))))
+
+(defn- temp-merge-worktree-path
+  [run-id task-id input-key]
+  (str (System/getProperty "java.io.tmpdir")
+       "/miniforge-merge/" run-id "/" task-id "/" input-key))
+
+(defn- enumerate-conflicts
+  "Parse `git ls-files --unmerged` output into `[{:path ... :stage ...}]`."
+  [host-repo worktree-path]
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 2a3a4cc1 — docstring now describes the per-path shape with the :stages vector accurately, and the unused host-repo argument is dropped. Call site updated.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,295 @@
...
+       :anomaly {:anomaly/category :anomalies/dag-multi-parent-conflict
+                 :anomaly/message  "Multi-parent merge conflicted"
+                 :task/id          task-id
+                 :merge/parents    parents
+                 :merge/conflicts  (enumerate-conflicts host-repo worktree-path)
+                 :merge/strategy   strategy
+                 :merge/input-key  input-key
+                 :git/exit-code    (:exit r)
+                 :git/stderr       (:err r)}})))
+
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 2a3a4cc1 — added a supported-merge-strategies guard at the top of merge-parent-branches!. Right now only :git-merge is supported; :sequential-merge ships in Stage 4. Plans that explicitly request an unsupported strategy now surface :anomalies/dag-multi-parent-strategy-unsupported with the requested strategy and the supported set echoed in the anomaly payload. New unsupported-strategy-anomaly-test pins the path.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,295 @@
...
+                                  {:merge/ok? true
+                                   :merge/ref ref-name
+                                   :merge/commit-sha (:commit-sha outcome)
+                                   :merge/input-key input-key
+                                   :merge/strategy strategy
+                                   :merge/parents parents})
+                              (do (cleanup-worktree! host-repo worktree)
+                                  (:anomaly outcome)))
+                      (seq (concat (:collapsed deduped) collapsed))
+                      (assoc :merge/collapsed (vec (concat (:collapsed deduped) collapsed))))))))))))))
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 2a3a4cc1 — added existing-merge-ref-sha that checks for the namespaced ref before doing the merge. If it already exists, return its SHA with :merge/cache-hit? true and skip the merge entirely. New idempotency-second-call-reuses-merge-ref-test runs merge-parent-branches! twice on the same inputs and asserts the second call returns the same SHA (not a fresh merge commit).

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,295 @@
...
+                  (let [outcome (run-merge! worktree host-repo task-id strategy parents input-key)]
+                    (cond-> (if (:ok? outcome)
+                              (do (run-git host-repo "update-ref" ref-name (:commit-sha outcome))
+                                  (cleanup-worktree! host-repo worktree)
+                                  {:merge/ok? true
+                                   :merge/ref ref-name
+                                   :merge/commit-sha (:commit-sha outcome)
+                                   :merge/input-key input-key
+                                   :merge/strategy strategy
+                                   :merge/parents parents})
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 2a3a4cc1 — update-ref's result is now captured; non-zero exit produces a typed :anomalies/dag-multi-parent-merge-failed with the ref name, commit-sha, git exit code, and stderr in the payload. The function will no longer return :merge/ok? true with a :merge/ref pointing nowhere.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -274,15 +574,17 @@
...
+   `acquire-environment!` forks off that branch instead of the spec
+   branch. `:branch` is ALWAYS set when task-def is given:
+   - Zero deps or single-dep unregistered → the spec branch.
+   - Single-dep registered → dep's branch (v1 chaining payoff).
+   - Multi-dep (v2) → the merged ref produced by `merge-parent-branches!`.
+     If the merge produces a typed anomaly (conflict / unrelated
+     histories / branch unresolvable), it propagates back to the caller
+     via `:dag/merge-anomaly` on opts; `execute-single-task` checks for
+     this and fails the task with the anomaly instead of running the
+     sub-workflow.
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 2a3a4cc1 — task-sub-opts docstring now describes the four cases (zero-dep, single-dep, multi-dep success, multi-dep anomaly) and explicitly calls out the :branch omission on the multi-dep anomaly path. The unit test continues to pin that contract.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -367,33 +683,45 @@
...
-        ;; it for downstream tasks' base resolution.
-        task-branch (assoc :task-branch task-branch))
-      (workflow-failure (extract-sub-workflow-error result) metrics))))
+        merge-anomaly (:dag/merge-anomaly sub-opts)]
+    (if merge-anomaly
+      ;; Multi-parent merge failed before we even entered the sub-workflow.
+      ;; Surface the anomaly directly; Stage 2 will replace conflicts with
+      ;; the resolution sub-workflow.
+      (workflow-failure (or (:anomaly/message merge-anomaly)
+                            "Multi-parent merge failed")
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 2a3a4cc1 — run-mini-workflow now passes the full typed anomaly through to workflow-failure as :error instead of stringifying it. workflow-result->dag-result was already shaped to accept any value at :error, so this lets downstream consumers — the dashboard, evidence bundle, and the Stage 2 resolution sub-workflow input — see :anomaly/category, :merge/conflicts, :merge/parents, :git/stderr in their structured form.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,366 @@
...
+           :git/exit-code (:exit r)
+           :git/stderr (:err r)}))
+      {:parents (persistent! out)})))
+
+(defn- ancestor-of?
+  "True when `sha-a` is an ancestor of (reachable from) `sha-b`."
+  [host-repo sha-a sha-b]
+  (zero? (:exit (run-git host-repo "merge-base" "--is-ancestor" sha-a sha-b))))
+
+(defn- collapse-ancestors
```
H31MDALLR, line 517: Factor out these anonymous functions. Except the one that you're using for memoize, that's small enough. The rest of them need to be in their own function at the right strata.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,366 @@
...
+     :collapsed collapsed}))
+
+(defn- shared-ancestry?
+  "True when at least one common ancestor exists across all parent SHAs.
+   Two-parent case is the common one; for 3+, git's merge-base requires
+   `--octopus` to find the n-way common ancestor."
+  [host-repo parents]
+  (let [shas (mapv :commit-sha parents)
+        args (cond-> ["merge-base"]
+               (> (count shas) 2) (conj "--octopus")
```
H31MDALLR: No magic numbers. What do this "2" actually mean?

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,366 @@
...
+         (not (str/blank? (:out r))))))
+
+(def ^:private merge-base-ref-prefix "refs/miniforge/dag-base")
+
+(defn- merge-base-ref-name
+  "Spec §7.2 step 3. Namespaced ref under `refs/miniforge/dag-base/...`
+   isolating the merge by run-id, task-id, and the input-key (so retries
+   of the same effective input reuse the same ref)."
+  [run-id task-id input-key]
+  (str merge-base-ref-prefix "/" run-id "/" task-id "/" input-key))
```
H31MDALLR, line 549: You don't have raw strings in Miniforge. You can move that to a system locale if it's never meant to be localized. Same for all the other raw strings in this file.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,366 @@
...
+
+   `git ls-files --unmerged` emits one line per stage entry per
+   conflicted path (typically stages 1/2/3 = base/ours/theirs); we
+   collapse to one entry per path with the observed stages so the
+   resolution sub-workflow (Stage 2) sees each conflicted path once
+   alongside which stages git surfaced for it."
+  [worktree-path]
+  (let [r (run-git worktree-path "ls-files" "--unmerged")
+        lines (when (zero? (:exit r))
+                (->> (str/split-lines (str (:out r)))
```
H31MDALLR:
```clojure
(->> r :out str str/splitlines (remove str/blank?))
```

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,366 @@
...
+                     (remove str/blank?)))]
+    (->> lines
+         (map (fn [line]
```
H31MDALLR: Extract both of these anonymous functions to a function at the appros strata.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,366 @@
...
+        merge-strategy (if (> (count parents) 2) "octopus" "ort")
+        merge-args (concat ["merge" "-s" merge-strategy]
+                           (pinned-merge-flags message)
+                           (map :commit-sha rest-parents))
+        r (apply run-git worktree-path merge-args)]
+    (cond
+      (zero? (:exit r))
+      (let [head (run-git worktree-path "rev-parse" "HEAD")]
+        (if (zero? (:exit head))
+          {:ok? true :commit-sha (str/trim (:out head))}
```
H31MDALLR, line 646: We seem to be introducing a different type of success, and we already have success in anomaly patterns that are factory patterns in miniforge. We should be using those.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,366 @@
...
+          {:ok? false
+           :anomaly {:anomaly/category :anomalies/dag-multi-parent-merge-failed
+                     :anomaly/message "Merge succeeded but rev-parse HEAD failed"
```
H31MDALLR: No raw strings. Localize or localize to system if it's never going to be seen.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,366 @@
...
+      ;; No deps registered — fall back to default branch (matches the
+      ;; single-parent fast-path's defensive semantics).
+      (not (seq (:merge/parents resolved)))
+      {:merge/ok? true
+       :merge/ref (default-spec-branch context)
+       :merge/single-parent? true
+       :merge/fallback-reason :no-registered-parents}
+
+      :else
+      (let [snapshot (snapshot-parent-shas host-repo (:merge/parents resolved))]
```
H31MDALLR: This is huge. I think this should be extracted and possibly even decomposed after it's extracted.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,366 @@
...
+      {:anomaly/category :anomalies/dag-multi-parent-strategy-unsupported
+       :anomaly/message  (str "Merge strategy " strategy " is not supported in this stage")
+       :task/id          task-id
+       :merge/strategy   strategy
+       :merge/supported  supported-merge-strategies}
+
+      ;; No deps registered — fall back to default branch (matches the
+      ;; single-parent fast-path's defensive semantics).
+      (not (seq (:merge/parents resolved)))
+      {:merge/ok? true
```
H31MDALLR: This looks to me like something that should be in a factory function that creates these maps because I bet you have more than one.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -266,6 +277,366 @@
...
+              (= 1 (count parents))
+              (cond-> {:merge/ok? true
+                       :merge/ref (:branch (first parents))
+                       :merge/commit-sha (:commit-sha (first parents))
+                       :merge/single-parent? true}
+                (seq all-collapsed) (assoc :merge/collapsed all-collapsed))
+
+              ;; No shared ancestry → unrelated-histories anomaly per §6.5.
+              (not (shared-ancestry? host-repo parents))
+              {:anomaly/category :anomalies/dag-multi-parent-unrelated-histories
```
H31MDALLR: At least the second instance of this map, create a factory function for these.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
```diff
@@ -297,19 +673,33 @@
...
+                       (nil? task-def)
+                       nil
+
+                       (dag/multi-parent? deps)
+                       (merge-parent-branches! context task-def)
+
+                       :else
+                       (resolve-task-base-branch context task-def))
+         resolved-branch (cond
+                           (string? base-result)            base-result
```
H31MDALLR: Why do we have a string or a map question here?

components/workflow/test/ai/miniforge/workflow/merge_parent_branches_integration_test.clj
```diff
@@ -0,0 +1,314 @@
...
+
+(def ^:dynamic *repo* nil)
+
+(defn- run-git!
+  "Run a git command in `cwd`. Throws on non-zero exit so test setup
+   bugs surface immediately rather than as cryptic downstream failures."
+  [cwd & args]
+  (let [r (apply shell/sh "git" "-C" cwd args)]
+    (when-not (zero? (:exit r))
+      (throw (ex-info (str "git " (str/join " " args) " failed: " (:err r))
```
H31MDALLR: Should be using exceptions as data. I am unsure we should be throwing here. Unless this is getting thrown right back to the user. If so, these strings need to be localized. They should be pulled out into a system or a user locale, either way, en-us or system.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
H31MDALLR: Fixed in 72dad83d. Extracted maximal-tip?, find-absorber, collapse-record from collapse-ancestors; parse-unmerged-line and summarize-conflicts-by-path from enumerate-conflicts; format-parent-line from deterministic-merge-message. Each is named at the appropriate stratum with a docstring describing its single responsibility. Kept the memoize closure as you specified — that one's small enough.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
H31MDALLR: Fixed in 72dad83d. The 2 was the threshold above which git's octopus strategy is required; now there's a named constant octopus-merge-min-parents (= 3) plus needs-octopus-strategy? and merge-strategy-name predicates so the call sites read 'is the parent count enough to need octopus' instead of '> 2'. Same for the merge-base path: merge-base-default-max-parents (= 2) names the threshold for when --octopus is needed there too.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
H31MDALLR: Fixed in 72dad83d. All anomaly messages and the merge commit message body now route through workflow.messages/t with new entries under :dag.merge/* in the workflow message catalog. Operator-internal strings (the test fixture's git failure message) are under :dag.merge.system/* — same routing, just marked system-locale so they're never user-translated but still have one place to audit and change.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
H31MDALLR: Fixed in 72dad83d. The chain now reads (->> r :out str str/split-lines (remove str/blank?)).

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
H31MDALLR: Fixed in 72dad83d. parse-unmerged-line (parses one line of git ls-files --unmerged output) and summarize-conflicts-by-path (collapses [{:path :stage}] to [{:path :stages [...]}]) are both extracted as named functions with docstrings.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
H31MDALLR: Fixed in 72dad83d. Replaced the bespoke {:merge/ok? true ...} success shape with dag/ok across the merge layer. Three typed factories now: merge-ok-result (full merge commit on namespaced ref), single-parent-fast-path-result (collapse-to-one), empty-registry-fallback-result (defensive). All wrap in dag/ok so consumers use the same dag/ok? / (:data ...) shape that the rest of the orchestrator uses for task results. task-sub-opts also lifts the v1 single-parent string return into dag/ok so downstream branching has a single shape.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
H31MDALLR: Fixed in 72dad83d. The strategy-unsupported message now goes through workflow.messages/t :dag.merge/strategy-unsupported with the strategy name as a parameter. Same approach as the rest of the v2 catalog entries.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
H31MDALLR: Fixed in 72dad83d. merge-parent-branches! decomposed into prepare-effective-parents (strategy validation + resolve + snapshot + dedupe + ancestor collapse + ancestry check; pure pipeline of registry + git, doesn't touch worktrees) and attempt-merge-with-cache! (cache check + worktree setup + merge + ref write + cleanup). merge-parent-branches! itself is now ~30 lines: strategy guard, prepare, branch on prepare result, attempt if needed.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
H31MDALLR: Fixed in 72dad83d. Six anomaly factories built — branch-unresolvable-anomaly, unrelated-histories-anomaly, strategy-unsupported-anomaly, conflict-anomaly, worktree-setup-failed-anomaly, ref-write-failed-anomaly. Each takes the minimum data and returns the canonical shape. No more inline {...} construction at the failure sites.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
H31MDALLR: Fixed in 72dad83d. Same factory family as the previous comment — all merge-anomaly construction now goes through the typed factories instead of inline maps. Three success-shape factories likewise (merge-ok-result, single-parent-fast-path-result, empty-registry-fallback-result), all wrapping in dag/ok.

components/workflow/src/ai/miniforge/workflow/dag_orchestrator.clj
H31MDALLR: Fixed in 72dad83d. The string-or-map duality is gone. task-sub-opts now consumes a single shape: success → dag/ok wrapping {:branch ...}, failure → anomaly map. The v1 single-parent path that previously returned a raw string is now lifted to dag/ok at the boundary, so downstream branching has one shape.

components/workflow/test/ai/miniforge/workflow/merge_parent_branches_integration_test.clj
H31MDALLR: Fixed in 72dad83d. The throw stays — test setup failures should be loud, and the test case will surface the cause regardless. But the message is now routed through workflow.messages/t under :dag.merge.system/git-test-failure (system-locale, never user-translated, but in the catalog so we have one place to audit dev-error wording). No more raw strings in the test fixture either.

PR #783, opened 2026-05-05
fix(reviewer): wire :prompt/progress-monitor + bump stagnation threshold

H31MDALLR:
## Summary

Stopgap: wires `:prompt/progress-monitor` into the reviewer agent (mirrors planner.clj/edn from PR #781), with a 180s stagnation threshold.

## Why

The 2026-05-04 `event-log-tool-visibility` dogfood (post #779/#781/#782) walked explore → plan → implement → verify successfully. Review #1 ran 4.3m with **4/4 quality gates passing** but came back `:rejected`. Inspecting the persisted review artifact:

```
:decision :rejected
:gates-passed 4 / :gates-failed 0
:blocking-issues ["Adaptive timeout: Stagnation timeout: no progress for 120119ms"]
```

Same class of bug as the planner stagnation we fixed in #781 — different agent. Reviewer's LLM call passed only `:system + :max-turns` to `llm/chat`, no `:progress-monitor`. Supervisor fell back to the framework default 120s, which is too tight for Opus on a review prompt that embeds the full implement artifact (8+ source files, 50–100k tokens) — the model spent 120s+ on Read/Grep before emitting structured EDN, supervisor killed it, runtime synthesized "Review rejected" with a stagnation-timeout blocking issue.

## Caveat — this is a stopgap

The threshold-bump pattern (60s → 180s for planner, now 120s → 180s for reviewer) is reactive. Each bump calibrates to whichever spec broke us last. Prompt size, model latency, tool-call depth — none of it is knowable from a spec EDN. The proper fix is a real progress detector that watches *what* the agent is doing (loop signatures, no-commitment patterns, repeated tool calls) rather than *whether stdout is alive*. Tracked separately as a follow-up spec; this PR is the threshold-bump that lets the next dogfood reach the next blocker.

## Changes

- `reviewer.edn`: add `:prompt/progress-monitor` block (180s stagnation, 600s max-total). Version 1.0.0 → 1.1.0.
- `reviewer.clj`: introduce `reviewer-prompt-data` delay + `create-reviewer-progress-monitor` (mirrors planner.clj). Wire monitor onto chat / chat-stream opts when config is present; falls back to framework default when EDN omits the block.

## Test plan

- [x] `bb test` — 5025 tests, 0 failures, 0 errors
- [ ] Re-run `event-log-tool-visibility` dogfood (resume from checkpoint) — expected: review completes without timing out

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #784, opened 2026-05-05
feat(progress-detector): stage 1 — detector framework + tool registry

H31MDALLR:
## Summary

Lands the substrate for semantic loop detection per the new spec at \`work/progress-detector-semantic-loop-detection.spec.edn\`. This is **Stage 1** of a three-stage rebuild of agent-supervision semantics — the bones-only landing that replaces wallclock-as-progress for within-turn detection.

**Stacked on PR #783** (reviewer stagnation threshold). Once #783 merges to main, this PR rebases against main.

## What lands

- New component \`components/progress-detector/\`
- **Detector protocol** (\`init\` / \`observe\` reducer contract) with \`NullDetector\` and \`MultiDetector\` reference implementations
- **Anomaly schema** — pure-description only. No \`:recommended-action\`, no \`:confidence\`. Detectors classify (\`:class\` ∈ {\`:mechanical\` \`:heuristic\`}, \`:severity\` ∈ {\`:info\` \`:warn\` \`:error\` \`:fatal\`}); the supervisor disposes (Stage 3 for full \`:on-anomaly\` map).
- **Tool capability registry** with per-tool \`:fingerprint/dimensions\` and \`:determinism\` (\`:stable-with-resource-version\` / \`:stable-ish\` / \`:environment-dependent\` / \`:unstable\`). Determinism gates whether a fingerprint repeat is mechanical or heuristic.
- **Layered config** (\`:inherit\` / \`:tune\` / \`:disable\` / \`:enable\` overlays with deep-merge semantics for \`:tune\`).
- 52 new tests under \`components/progress-detector/test\` exercising the framework end-to-end.

## What does NOT land (intentionally — see \`:spec/deferred\`)

- **Stage 2**: full normalized event envelope, \`:resource/version-hash\` on every tool result, progress ledger reducer, \`:detector/tool-loop\`, \`:detector/no-information-gain\`, \`:detector/repeated-failure\`.
- **Stage 3**: phase contracts, phase status taxonomy, full supervisor \`:on-anomaly\` policy, composite-rules for promoting heuristics, \`:detector/text-similarity\`, workflow-runner enforcement.

Stage 1 is the substrate. Stage 2 plugs detectors into it. Stage 3 plugs phase contracts + the policy supervisor on top.

## Provenance

This commit bundles autonomous-dogfood output (workflow \`8032a5fa\`) recovered from the persisted bundle, with fixes for every issue review #1 flagged:

| Reviewer issue | Fix |
|---|---|
| \`config.clj/merge-config\` drops base-layer params on implicit \`:inherit\` | Base layer's \`:config/params\` now seed the accumulator; subsequent layers apply via \`apply-directive\` normally |
| \`interface_test.clj\` lifecycle test puts state in detector slot via mistaken \`->\` thread | Replaced with explicit \`let\`-binding; preserves the protocol's \`[detector state observation]\` order needed for \`(partial observe det)\` reducer use |
| \`observe\` docstring contract unclear (multi-arity wrapper) | Documented argument order, immutability, no-throw, and the threading caveat |
| deps.edn indentation drift | Pre-existing (col-22 entries below \`dag-executor\` are the outlier); deferred — non-blocking warning |

## Commit budget

The commit uses \`MINIFORGE_COMMIT_BUDGET_OVERRIDE\` per the hook's override protocol. Rationale: this is a recovery commit of a 2,144-line autonomous-dogfood output as a single coherent component landing. Every individual ns is itself >100 reportable LOC (interface.clj alone is 235), so any meaningful slice exceeds the 200-LOC threshold; splitting across commits would land non-compiling intermediate states (\`protocol.clj\` defines the \`Detector\` protocol that every other ns consumes).

## Test plan

- [x] \`bb test\` — 5077 / 0 / 0 (5025 pre-existing + 52 new)
- [ ] When integrated with Stage 2's tool-loop detector, replay the recorded \`event-log-tool-visibility\` event sequence through the framework and confirm <5ms-per-event detector overhead via criterium


H31MDALLR:
## Standards-compliance review pass — `ad9460fb`

Reviewed the recovered autonomous-dogfood output against `components/phase/resources/packs/miniforge-standards.pack.edn`. Three classes of issue found:

**Real bug (mechanical):**
- `tool_profile.clj/validate-all` was calling `schema/explain-anomaly` on a tool-profile — humanizing errors against the wrong schema. Errors would have cited Anomaly fields (\`:anomaly/category\`, \`:anomaly/severity\`, …) instead of ToolProfile fields. Added `explain-tool-profile` + `explain-detector-config` to `schema.clj` and routed `validate-all` through the right one. Re-exported via `interface.clj`.

**Doc/impl drift:**
- `tool_profile.clj/load-registry` docstring promised "schema validation and warnings for invalid profiles" — the impl only checked `(map? raw)`. Rewrote the docstring to match impl, point callers at `validate-all` for per-entry validation, and explain why we don't validate at load time (one bad entry must not take the whole registry offline).

**Convention drift on `:std/stratified-design`:**
- 482 sibling source files in `components/*` use the canonical `;------ Layer N` markers. The autonomous output used a `;; ----` separator with no Layer label across 10/10 files. Converted all 60+ separators to canonical form via batch script; no behavior change.

## The interesting finding

The autonomous agent had access to the standards pack via context during planning AND review. It still produced output that drifted on every file. The reviewer didn't catch any of these — they reviewed for code-correctness (caught the protocol-dispatch and config-merge bugs in review #1) but didn't audit against the standards pack.

That's a real signal for the **Stage 3** spec work — `:phase/contract` shouldn't just check abstract layering principles; it needs to validate against the pack's example markers (regex over comment headers, schema-explainer fn-name spelling matches, etc.). Would have caught all three classes here.

## Test plan

- [x] `bb test` — 5077 / 0 / 0 (no behavior change beyond the explain-tool-profile fix)
- [ ] Sibling-file consistency: `grep -L ';------------------------------------------------------------------------------ Layer' components/progress-detector/src/**/*.clj` returns empty

components/progress-detector/src/ai/miniforge/progress_detector/interface.clj
```diff
@@ -0,0 +1,381 @@
...
+     config       - config map passed to init
+     observations - seq of observation maps
+
+   Returns: Final state after all observations are folded."
+  proto/reduce-observations)
+
+;------------------------------------------------------------------------------ Layer 1
+;; Layer 1 — Anomaly schema (re-exported)
+
+(def valid-anomaly?
```
H31MDALLR: Shouldn't validity be owned by the anomaly component?

components/progress-detector/src/ai/miniforge/progress_detector/interface.clj
```diff
@@ -0,0 +1,381 @@
...
+
+(def explain-tool-profile
+  "Return humanized error map if m fails ToolProfile validation, else nil."
+  schema/explain-tool-profile)
+
+(def valid-detector-config?
+  "Return true if m satisfies the DetectorConfig schema."
+  schema/valid-detector-config?)
+
+(def explain-detector-config
```
H31MDALLR: I feel like these should be owned by the components that vendor that functionality. That way we can simply have a registration process for things to register with the progress detector, and not have to worry about the progress detector knowing about all the things that it is tracking progress for.

components/progress-detector/src/ai/miniforge/progress_detector/interface.clj
```diff
@@ -0,0 +1,381 @@
...
+   (tp/detector-kind-of tool-id registry)))
+
+(def all-tool-ids
+  "Return sorted vector of all registered tool-id keywords.
+
+   Arguments:
+     registry - (optional) registry map"
+  tp/all-tool-ids)
+
+(defn register-tool
```
H31MDALLR: We already have a tool registration and a lot of tool component-level stuff in Miniforge. Make sure we're not duplicating most of that work over there. It feels to me like the tool registry already provides 90% of this, and really these are just additive things perhaps (if there is a delta at all) for MCP tools.

components/progress-detector/src/ai/miniforge/progress_detector/interface.clj
H31MDALLR: Fixed in 05d81bc7. progress-detector now declares ai.miniforge/anomaly as a deps.edn dependency and re-exports anomaly's Anomaly schema + anomaly? predicate. Detector-specific structure (:detector/kind, :anomaly/class, :anomaly/severity, :anomaly/category, :anomaly/evidence) lives inside :anomaly/data via a new DetectorAnomalyData schema. The duplicate Anomaly definition is gone.

components/progress-detector/src/ai/miniforge/progress_detector/interface.clj
```diff
@@ -0,0 +1,381 @@
...
+   Returns: Vector of anomaly maps (possibly empty)."
+  [state]
+  (get state :anomalies []))
+
+(defn anomalies-by-severity
+  "Return anomalies from state filtered to the given severity.
+
+   Arguments:
+     state    - detector state map
+     severity - :anomaly.severity/critical | :warning | :info
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 05d81bc7. Severity vocabulary aligned with the spec's :info|:warn|:error|:fatal (the autonomous output had invented :anomaly.severity/critical|warning|info). The helper now reads severity from [:anomaly/data :anomaly/severity] since severity is detector metadata, not part of the canonical anomaly shape. critical-anomalies is gone; fatal-anomalies replaces it.

components/progress-detector/src/ai/miniforge/progress_detector/interface.clj
H31MDALLR: Fixed in 05d81bc7. tool_profile.clj is now a thin registration API — make-registry / register! / unregister! / lookup. The EDN-driven default-registry is gone (it had no real owner). Stage 1 ships an empty default-registry; Stage 2 wires adapter-claude-code, agent, etc. to register their profiles at namespace-load time. progress-detector no longer carries first-party knowledge of which tools exist or what their characteristics are.

components/progress-detector/src/ai/miniforge/progress_detector/interface.clj
H31MDALLR: Acknowledged. The fingerprint/determinism delta over tool-registry's existing tool config is small enough that Stage 2 should consider folding it into tool-registry's tool config schema rather than maintaining a parallel registry. Recorded in the spec as a Stage 2 question. Stage 1 ships only the registration mechanism, not parallel tool data — components that vendor tools own their profiles.

work/progress-detector-semantic-loop-detection.spec.edn
```diff
@@ -0,0 +1,395 @@
...
+  ## Component layout
+
+  New: components/progress-detector/
+    src/ai/miniforge/progress_detector/
+      interface.clj          ;; public Layer 0/1 boundary
+      protocol.clj           ;; Detector protocol (reducer shape)
+      anomaly.clj            ;; anomaly schema + ctor
+      tool_profile.clj       ;; tool capability registry
+      detectors/
+        tool_loop.clj        ;; :detector/tool-loop
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 05d81bc7. Spec component-layout section now describes Stage 1 actual reality (interface.clj, protocol.clj, schema.clj, tool_profile.clj, config.clj). Files that don't ship in Stage 1 (anomaly.clj, detectors/tool_loop.clj, detectors/repair_loop.clj, supervisor.clj) are explicitly called out as deferred + which stage owns them. resources/.../tool-profiles.edn removed entirely per the user-review on tool-registry duplication.

components/progress-detector/src/ai/miniforge/progress_detector/config.clj
```diff
@@ -0,0 +1,240 @@
...
+        ;; accumulator. Without this seeding the base's params get
+        ;; dropped on its own implicit :inherit directive (see
+        ;; merge-config-tune-test, overlay-test, deep-merge-nested-test
+        ;; — all surfaced this when the base layer's params went
+        ;; missing from the resolved config). Subsequent layers apply
+        ;; via apply-directive normally.
+        [base-layer & overlay-layers] layers
+        seed (-> {:detector/enabled? (get base-layer :detector/enabled? true)
+                  :config/params     (deep-merge {} (get base-layer :config/params {}))
+                  :config/directives [(resolve-directive base-layer)]})
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 05d81bc7. merge-config docstring now states explicitly: directives apply to overlay layers only; the base layer's :config/directive is recorded in :config/directives for audit but never executed (a base :disable would be self-defeating). Existing tests hold.

work/progress-detector-semantic-loop-detection.spec.edn
```diff
@@ -0,0 +1,395 @@
...
+
+  ## Anomaly schema (Layer 3)
+
+  Detectors emit pure description, no lifecycle:
+
+    {:anomaly/id          uuid?
+     :anomaly/category    keyword?           ;; :anomalies.agent/tool-loop, …
+     :anomaly/class       #{:mechanical :heuristic}
+     :anomaly/severity    #{:info :warn :error :fatal}
+     :detector/kind       keyword?
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 05d81bc7. Implementation now matches the spec's original vocabulary: :anomaly/class ∈ {:mechanical :heuristic} and :anomaly/severity ∈ {:info :warn :error :fatal}. The autonomous output's invented :anomaly.class/hard|soft|advisory and :anomaly.severity/critical|warning|info enums are gone. DetectorAnomalyData schema enforces the spec's vocabulary directly.

components/progress-detector/src/ai/miniforge/progress_detector/protocol.clj
```diff
@@ -0,0 +1,176 @@
...
+
+  (observe [_detector state observation]
+    (let [sub-states' (mapv #(observe %1 %2 observation)
+                            detectors
+                            (:sub-states state))
+          new-anomalies (into [] (mapcat :anomalies) sub-states')]
+      (-> state
+          (update :observations conj observation)
+          (assoc  :sub-states sub-states')
+          (update :anomalies into new-anomalies)))))
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 05d81bc7. MultiDetector now tracks per-sub-state anomaly counts before observe runs, then appends only the delta via subvec. The full :anomalies-history-on-every-step bug is gone; growth is O(N) in observations rather than O(N²).

PR #789, opened 2026-05-06
fix(review): normalize timeout-only reviewer failures

H31MDALLR:
## Summary

This PR packages the remaining Claude dogfood review-boundary fix from the `behavioral-verification-monitor` run on
`main`.

The remaining failure shape was narrower:

1. the reviewer could return a parsed review artifact with all gates passing
2. the only blocking issue could be an adaptive-timeout/stagnation message
3. Miniforge could still treat that as a real rejected review and redirect implement

### Reviewer timeout-only failures now fail as backend errors

- broaden the reviewer timeout-only classifier so it does not depend on the backend wrapper’s `success?` bit
- when the parsed review artifact says:
  - all gates passed
  - no real findings or recommendations exist
  - the only blocking issues are timeout/stagnation messages
  - the decision is negative
  then review now returns `:reviewer/backend-timeout` instead of a fake rejected review
- preserve the timeout message itself as the phase error message
- add regressions for both shapes:
  - backend wrapper reports failure
  - backend wrapper reports nominal success but the parsed review is still only a timeout echo

## Validation

- `clj-kondo` on touched files: clean
- focused JVM regression run:
  - `ai.miniforge.agent.reviewer-test`

## Live Behavior

- persisted dogfood review snapshot showed `4/4` gates passed, no real findings, and only an adaptive-timeout blocking
  issue
- that exact shape is now classified as a reviewer/backend failure instead of a code-review rejection

## Follow-up

- rerun the Claude dogfood workflow from the updated branch state so the new reviewer timeout fix is actually in play

### DIFF
--- a/components/agent/src/ai/miniforge/agent/reviewer.clj
+++ b/components/agent/src/ai/miniforge/agent/reviewer.clj
@@ -571,6 +571,41 @@
   [logger data]
   (log/info logger :reviewer :reviewer/phase-completed {:data data}))
 
+(defn- timeout-only-error-result
+  "Normalize the reviewer exit path when the LLM only reports its own timeout.
+
+   This preserves backend timeout metadata, emits the standard phase-completed
+   telemetry, and reports the deterministic gate outcome instead of converting
+   the backend failure into a bogus code-review rejection."
+  [logger normalized llm-review gate-result counts duration tokens cost-usd timeout-failure-message]
+  (log/warn logger :reviewer :reviewer/backend-timeout-only
+            {:data {:llm-decision (:review/decision llm-review)
+                    :gate-decision (:decision gate-result)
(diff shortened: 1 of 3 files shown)


components/agent/src/ai/miniforge/agent/reviewer.clj
```diff
@@ -649,23 +685,54 @@
...
-                                  :gates-passed (:passed counts)
-                                  :gates-failed (:failed counts)
-                                  :llm-issues (count llm-issues)
-                                  :duration-ms duration}})
-
-                ;; Phase lifecycle: mark review exit with decision
-                (leave-review logger {:review/decision final-decision
-                                      :duration-ms duration
+                (if timeout-only-review?
+                  (do
```
H31MDALLR: This feels like something we might want to extract into a function. Is this reusable?

PR #790, opened 2026-05-06
fix(cli): harden dogfood startup, no-op handoff, and resume

H31MDALLR:
## Summary

This PR packages the Claude dogfood fixes that are not review-policy specific:

1. short backend/version probes in CLI startup could still behave differently from the real inherited Claude process
2. a valid `:already-implemented` implement result could still be overwritten by curator `:no-files-written`
3. workflow resume could still die when a checkpoint recorded a synthetic DAG task workflow id instead of a loadable
  top-level workflow type

### CLI startup probes are supervised directly

- replace the short probe path in `workflow_runner.clj` with direct `ProcessBuilder` execution
- read stdout/stderr concurrently
- close stdin explicitly
- enforce timeout with forcible process teardown
- add focused probe tests for:
  - fast successful command capture
  - bounded timeout behavior

### Implement preserves verified no-op outcomes

- in `enter-implement`, a normalized `:already-implemented` result now survives curator `:curator/no-files-written`
- this keeps verified “task already satisfied” outcomes from being collapsed into terminal empty-diff failures
- existing repair-loop hardening remains in place:
  - unsupported `:already-implemented` after prior review/verify failures still fails closed
- add a regression covering the full `enter`/`leave` path for a legitimate already-implemented no-op

### Resume ignores synthetic DAG task workflow ids

- `workflow-resume` now treats keywords like `:dag-task-...` as internal synthetic workflow ids, not loadable workflow
  types
- identity resolution now prefers:
  - recorded workflow spec
  - non-synthetic machine snapshot workflow id
  - fallback selection profile
- this fixes resume runs that previously died with `Workflow not found` while replaying task-subworkflow checkpoints
- add a regression for the synthetic DAG-task id shape

## Validation

- `clj-kondo` on touched files: clean
- focused JVM regression run:
  - `ai.miniforge.phase-software-factory.implement-test`
  - `ai.miniforge.workflow-resume.core-test`
  - `ai.miniforge.cli.workflow-runner.preflight-test`

## Live Behavior

- Claude dogfood on this branch now stamps `/Users/chris/.local/bin/claude 2.1.129`
- `:explore` passes
- `:plan` passes
- the first DAG implement/verify cycle gets past startup and no-op handoff issues

### DIFF
--- a/components/agent/src/ai/miniforge/agent/artifact_session.clj
+++ b/components/agent/src/ai/miniforge/agent/artifact_session.clj
@@ -506,6 +506,28 @@
                        (.getPath f) "—" (ex-message e)))
             nil))))))
 
+(defn read-capsule-worktree-artifact
+  "Read a role artifact from <workdir>/.miniforge/<role>.edn inside a capsule.
+
+   Returns the parsed artifact map or nil. Missing files and parse failures are
+   treated as non-fatal, matching host-mode read-worktree-artifact behavior."
+  [session role]
+  (when (and session role)
+    (let [path (str (:workdir session) "/.miniforge/" (name role) ".edn")
+          result ((:exec! session) (:executor session) (:environment-id session)
+                  (str "cat " path) {:workdir (:workdir session)})
+          content (get-in result [:data :stdout] "")]
+      (when (seq content)
+        (try
+          (-> content
+              edn/read-string
+              parse-uuid-strings)
+          (catch Exception e
+            (binding [*out* *err*]
+              (println "WARN: failed to parse capsule worktree artifact at"
+                       path "—" (ex-message e)))
+            nil))))))
+
 (defn read-artifact
   "Read the artifact EDN file from a session directory.
 
@@ -761,10 +783,14 @@
   (try
     (let [result (body-fn session)
           workdir (:workdir session)
-          worktree-artifacts (when (and (= :host mode) workdir)
+          read-role-artifact (case mode
+                               :capsule #(read-capsule-worktree-artifact session %)
+                               :host    #(read-worktree-artifact workdir %)
+                               (constantly nil))
+          worktree-artifacts (when workdir
                                (into {}
                                      (keep (fn [role]
-                                             (when-let [a (read-worktree-artifact workdir role)]
+                                             (when-let [a (read-role-artifact role)]
                                                [role a])))
                                      [:plan :implement :verify :review :release]))]
       {:llm-result result
--- a/components/agent/src/ai/miniforge/agent/artifact_session.clj
+++ b/components/agent/src/ai/miniforge/agent/artifact_session.clj
@@ -48,6 +48,23 @@
    [:mcp-config-path [:string {:min 1}]]
    [:artifact-path [:string {:min 1}]]])
 
+(def ^:private system-message-templates
+  "System-localized stderr message templates for artifact-session I/O."
+  {:warn/worktree-artifact-parse
+   "WARN: failed to parse worktree artifact at %s — %s"
+
+   :warn/capsule-worktree-artifact-parse
+   "WARN: failed to parse capsule worktree artifact at %s — %s"
+
+   :error/artifact-file-missing
+   "ERROR: artifact file not found at %s — MCP tool was likely not called by the LLM"
+
+   :warn/artifact-parse
+   "WARN: failed to parse artifact at %s — %s"
+
+   :warn/context-misses-parse
+   "WARN: failed to parse context misses at %s — %s"})
+
 (defn validate-session
   "Validate a session map against the Session schema.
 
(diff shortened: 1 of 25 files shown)


components/agent/src/ai/miniforge/agent/artifact_session.clj
```diff
@@ -506,6 +506,28 @@
...
             nil))))))
 
+(defn read-capsule-worktree-artifact
+  "Read a role artifact from <workdir>/.miniforge/<role>.edn inside a capsule.
+
+   Returns the parsed artifact map or nil. Missing files and parse failures are
+   treated as non-fatal, matching host-mode read-worktree-artifact behavior."
+  [session role]
+  (when (and session role)
+    (let [path (str (:workdir session) "/.miniforge/" (name role) ".edn")
```
H31MDALLR: Should these be essentially templated so that we don't hand-compose the string parts in multiple places?

components/agent/src/ai/miniforge/agent/artifact_session.clj
```diff
@@ -506,6 +506,28 @@
...
+                  (str "cat " path) {:workdir (:workdir session)})
+          content (get-in result [:data :stdout] "")]
+      (when (seq content)
+        (try
+          (-> content
+              edn/read-string
+              parse-uuid-strings)
+          (catch Exception e
+            (binding [*out* *err*]
+              (println "WARN: failed to parse capsule worktree artifact at"
```
H31MDALLR: Make sure we're localized. System or user.

components/agent/src/ai/miniforge/agent/tester.clj
```diff
@@ -333,65 +334,61 @@
...
-                                               :cost-usd cost-usd}}))
-                    ;; LLM returned content but no parseable tests — fail explicitly
-                    (response/error "LLM response could not be parsed as test artifact"
-                                    {:tokens tokens})))
-                ;; LLM call failed — propagate error, no fallback
-                (response/error (or (:message (llm/get-error response))
-                                    "LLM call failed"))))
+              (if (result-boundary/usable-content? normalized)
+                (if-let [tests (result-boundary/authoritative-payload normalized)]
+                  (let [tests-with-meta (-> tests
```
H31MDALLR: This is doing a lot of work. That makes me think this should actually be pulled out and decomposed into its own utility function.

components/agent/test/ai/miniforge/agent/artifact_session_capsule_test.clj
```diff
@@ -40,9 +40,16 @@
...
   (fn [_executor _env-id command & [_opts]]
     (swap! log conj command)
-    (if (and (string? command) (str/starts-with? command "cat ") (str/includes? command "artifact.edn"))
+    (cond
+      (and (string? command) (str/starts-with? command "cat ") (str/includes? command "artifact.edn"))
       {:ok? true :data {:stdout "{:code/id \"a1b2c3d4-e5f6-7890-abcd-ef1234567890\" :code/description \"test\"}"
                         :stderr "" :exit-code 0}}
+
+      (and (string? command) (str/starts-with? command "cat ") (str/includes? command "implement.edn"))
+      {:ok? true :data {:stdout "{:status :already-implemented :summary \"already there\"}"
```
H31MDALLR: Use our standardized success and anomaly functions so that we don't drift from the standard.

components/agent/src/ai/miniforge/agent/artifact_session.clj
H31MDALLR: Templated this path. The artifact-session read warnings now go through a shared `system-message-templates` map plus `emit-system-message!`, and the role-path composition is centralized in `artifact-filename` / `worktree-artifact-file` instead of being hand-composed at each callsite.

components/agent/src/ai/miniforge/agent/artifact_session.clj
H31MDALLR: Localized as system messages. These stderr diagnostics now render from the shared system template map, so we are no longer assembling ad hoc operator-facing strings inline in the readers.

components/agent/src/ai/miniforge/agent/tester.clj
H31MDALLR: Decomposed this block. The test-artifact derivation now goes through `code-block-test-artifact`, `enrich-test-artifact`, `test-artifact-metrics`, and the two count helpers, so the main tester path only wires the boundary and response assembly.

components/agent/test/ai/miniforge/agent/artifact_session_capsule_test.clj
H31MDALLR: Switched the capsule test over to the standard DAG executor result constructors. It now uses `ai.miniforge.dag-executor.result/ok`, and I also cleaned up the test constants so the touched test code is no longer carrying repeated magic values.

PR #810, opened 2026-05-08
feat(event-stream): Snowflake event-id generator + filename grammar (BD-2b sub-1)

H31MDALLR:
## Summary

First implementation slice off the [BD-2 design RFC (#795)](https://github.com/miniforge-ai/miniforge/pull/795). Establishes the on-disk shape that BD-2b sub-2/sub-3 and BD-2c build on, kept tight enough to land on its own.

- **Snowflake generator** (new `event-stream.snowflake` namespace): 41-bit ms ts from a fixed miniforge epoch + 10-bit worker + 12-bit per-ms sequence. Encoded into a `java.util.UUID`'s most-significant 64 bits with low 64 zero — `:event/id` stays `uuid?` for downstream code while the leading hex sorts lexically by creation order. Worker-id lease via `FileLock` (auto-released on JVM exit). Clock-rollback stall + pre-epoch guard.
- **Integration in `event-stream.core`**: `create-event-stream` accepts `:snowflake-generator`; `create-envelope` calls it for `:event/id` when present, falls back to `random-uuid` otherwise. No contract change for downstream — `(uuid? :event/id)` continues to hold.
- **Filename grammar in `event-stream.sinks`**: `{event-id-hex16}__{workflow-seq-dec12}.transit.json` for snowflake-encoded events; legacy `{timestamp}-{uuid}.json` fallback for non-snowflake streams. Reader filter (`.endsWith ".json"`) accepts both.
- **16 new snowflake tests** + 2 new sinks tests pin the contract.

## Drive-by fixes (surfaced by this PR)

1. **Test-runner brick isolation.** `pr-sync` and `agent` tests `with-redefs` `clojure.java.shell/sh` to local mocks. The runner runs bricks in parallel within a single JVM, so those mocks bleed into other bricks' real subprocess calls — notably `workflow/merge-resolution-test`'s `git init`, where the mock returns `{:err "no route matched"}` or `{:err "fatal: not a git repo"}` depending on which mock is active. Added `pr-sync` and `agent` to `default-isolated-bricks` so each gets its own JVM. Deeper fix (mocking through private wrappers instead of vendor namespaces) is a multi-file refactor, deferred.
2. **Babashka compatibility.** `event-stream` is transitively loaded from BB-runtime bricks. `java.nio.channels.FileLock` and `java.lang.management` aren't in BB. Replaced top-level `:import` with FQN class references inside fn bodies so the namespace loads in BB; worker-id leasing only ever runs on the JVM. Dropped `pid`/`host` from lease metadata (informational; FileLock OS-release on JVM exit is what guarantees correctness).

## Out of scope (tracked for follow-on sub-PRs)

- BD-2b sub-2: Manifest schema + lease liveness model.
- BD-2b sub-3: Atomic archive + scheduled cleanup with budgets.
- BD-2c: Active checkpoints + snapshots + consumer hydration.
- CLI wiring of the generator into the production stream startup. `create-event-stream` accepts a generator but the CLI doesn't yet pass one — production events still use random-uuid `:event/id` and the legacy filename until the wiring lands. Migration is producer-opt-in via the `create-event-stream` seam.

## Validation

```
bb pre-commit
```

5388 tests, 23980 passes, 0 failures, 0 errors. GraalVM/Babashka compatibility green.

## Test plan

- [ ] reviewer agrees with the `:event/id`-as-snowflake-encoded-UUID design (preserves `uuid?` for downstream)
- [ ] reviewer agrees the brick-isolation drive-by is the right interim until the deeper private-wrapper refactor
- [ ] reviewer agrees BB compat via FQN class references is reasonable vs marking the brick `:jvm-only`

### DIFF
--- a/components/event-stream/src/ai/miniforge/event_stream/core.clj
+++ b/components/event-stream/src/ai/miniforge/event_stream/core.clj
@@ -20,6 +20,7 @@
   "Event bus and event constructors for workflow observability."
   (:require
    [ai.miniforge.event-stream.messages :as messages]
+   [ai.miniforge.event-stream.snowflake :as snowflake]
    [ai.miniforge.phase.interface :as phase]
    [ai.miniforge.logging.interface :as log]
    [ai.miniforge.response.interface :as response]
@@ -54,12 +55,22 @@
 ;; Event envelope constructor
 
 (defn create-envelope
-  "Create an event envelope with sequence numbering."
+  "Create an event envelope with sequence numbering.
+
+   When the stream carries a `:snowflake-generator` (BD-2b), the
+   envelope's `:event/id` is a snowflake-encoded UUID — the
+   most-significant 64 bits encode (epoch-ms timestamp, worker id,
+   per-ms sequence) per the RFC, so events sort lexically by creation
+   order. Streams without a generator fall back to a random UUID;
+   `:event/id` stays a `uuid?` either way for downstream compatibility."
   [stream event-type workflow-id message]
-  (let [seq-num (get-in @stream [:sequence-numbers workflow-id] 0)]
+  (let [seq-num (get-in @stream [:sequence-numbers workflow-id] 0)
+        generator (:snowflake-generator @stream)]
     (swap! stream assoc-in [:sequence-numbers workflow-id] (inc seq-num))
     {:event/type event-type
-     :event/id (random-uuid)
+     :event/id (if generator
+                 (snowflake/next-id! generator)
+                 (random-uuid))
      :event/timestamp (java.util.Date.)
      :event/version event-version
      :event/sequence-number seq-num
@@ -87,7 +98,10 @@
      (create-event-stream {:sinks [(sinks/file-sink) (sinks/stdout-sink)]})
 
      ;; From config
-     (create-event-stream {:config user-config})"
+     (create-event-stream {:config user-config})
+
+     ;; With a Snowflake event-id generator (BD-2b)
+     (create-event-stream {:snowflake-generator (snowflake/create-generator)})"
   [& [opts]]
   (let [;; Create sinks from config or use provided sinks or default
         event-sinks (cond
@@ -108,7 +122,11 @@
            ;; reaching zero before draining sinks. `publish!` increments
            ;; on entry (after the quiesce check) and decrements in a
            ;; finally so a sink exception still releases the slot.
-           :in-flight 0})))
+           :in-flight 0
+           ;; BD-2b: optional Snowflake event-id generator. When present,
+           ;; `create-envelope` calls it for `:event/id` so events sort
+           ;; lexically by creation order. nil = random-uuid fallback.
+           :snowflake-generator (:snowflake-generator opts)})))
 
 ;------------------------------------------------------------------------------ Layer 0
 ;; publish! helpers — small, single-purpose pieces composed by publish!
--- a/components/event-stream/src/ai/miniforge/event_stream/core.clj
+++ b/components/event-stream/src/ai/miniforge/event_stream/core.clj
@@ -84,9 +84,16 @@
   "Create an event stream with configurable sinks.
 
    Options:
-     :logger - Optional logger instance
-     :sinks - Vector of sink functions (default: file sink)
-     :config - Config map to create sinks from
+     :logger              - Optional logger instance
+     :sinks               - Vector of sink functions (default: file sink)
+     :config              - Config map to create sinks from
+     :snowflake-generator - Optional snowflake event-id generator
+                            (BD-2b). When supplied, `create-envelope`
+                            uses it for `:event/id` so events sort
+                            lexically by creation order. Without it,
+                            `:event/id` falls back to `random-uuid`
+                            and the file sink uses the legacy
+                            `{timestamp}-{uuid}.json` filename.
 
    Returns: Event stream atom
 
--- a/components/event-stream/src/ai/miniforge/event_stream/sinks.clj
+++ b/components/event-stream/src/ai/miniforge/event_stream/sinks.clj
@@ -20,13 +20,18 @@
   "Configurable event sinks for different deployment scenarios.
 
    Supported sinks:
-   - :file   - Write to ~/.miniforge/events/<workflow-id>/<timestamp>-<uuid>.json (local dev)
+   - :file   - Write to ~/.miniforge/events/<workflow-id>/<eventid>__<seq>.transit.json
+               (local dev). When the event's :event/id is a Snowflake-encoded UUID,
+               the leading hex of the filename sorts by creation order. Legacy
+               files written before BD-2b retain the {timestamp}-{uuid}.json
+               name and are still readable.
    - :stdout - Print to stdout (container/Docker/K8s)
    - :stderr - Print to stderr (error-only events)
    - :fleet  - Send to fleet command (org-level ops)
    - :multi  - Combine multiple sinks"
   (:require
    [ai.miniforge.config.interface :as config]
+   [ai.miniforge.event-stream.snowflake :as snowflake]
    [clojure.java.io :as io]
    [clojure.string :as str]
    [clojure.pprint :as pprint]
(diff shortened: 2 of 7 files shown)


components/event-stream/src/ai/miniforge/event_stream/snowflake.clj
```diff
@@ -0,0 +1,315 @@
...
+   [clojure.java.io :as io]
+   [clojure.string :as str]
+   [ai.miniforge.logging.interface :as log]))
+
+;; NOTE: Class references for the worker-id lease path
+;; (java.io.RandomAccessFile, java.nio.ByteBuffer, java.nio.channels.FileChannel,
+;; java.nio.channels.FileLock, java.lang.management.ManagementFactory,
+;; java.net.InetAddress, java.io.File) are used via fully-qualified names inside
+;; fn bodies rather than imported at the top, so the namespace loads in
+;; Babashka (which doesn't ship java.nio.channels.FileLock). Functions that
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 3ee5177. Trimmed the NOTE to only mention the classes the namespace actually still references (RandomAccessFile, ByteBuffer, FileChannel, FileLock). The dropped pid/host metadata is what removed the dependence on ManagementFactory/InetAddress/File.

components/event-stream/src/ai/miniforge/event_stream/snowflake.clj
```diff
@@ -0,0 +1,315 @@
...
+    (try
+      (let [lock (.tryLock channel)]
+        (if lock
+          (do
+            (write-lease-metadata! channel id)
+            {:channel channel :lock lock :worker-id id :raf raf})
+          (do (.close raf) nil)))
+      (catch Exception _
+        (.close raf)
+        nil))))
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 3ee5177. `try-acquire-slot` now catches `Exception` but discriminates inside the body via `(instance? (Class/forName "java.nio.channels.OverlappingFileLockException") e)`. Only the in-process overlapping-lock case is treated as "slot held"; everything else (IO, permission, etc.) is wrapped in an ex-info carrying `:reason :lease-io-failure`, `:worker-id`, `:lease-path`, `:workers-dir` and propagated, so a real failure doesn't masquerade as `:workers-exhausted`. The `Class/forName` indirection is intentional — a typed `(catch java.nio.channels.OverlappingFileLockException ...)` clause forces compile-time resolution and breaks Babashka load (the BB-runtime test transitively requires this namespace).

components/event-stream/src/ai/miniforge/event_stream/sinks.clj
```diff
@@ -20,13 +20,18 @@
   "Configurable event sinks for different deployment scenarios.
 
    Supported sinks:
-   - :file   - Write to ~/.miniforge/events/<workflow-id>/<timestamp>-<uuid>.json (local dev)
+   - :file   - Write to ~/.miniforge/events/<workflow-id>/<eventid>__<seq>.transit.json
+               (local dev). When the event's :event/id is a Snowflake-encoded UUID,
+               the leading hex of the filename sorts by creation order. Legacy
+               files written before BD-2b retain the {timestamp}-{uuid}.json
+               name and are still readable.
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 3ee5177. The `:file` bullet now spells out both filename shapes and the discriminator (snowflake-encoded UUID with low-64-bits-zero → new grammar; otherwise → legacy timestamp-leading), and notes both shapes coexist with the reader accepting either.

components/event-stream/src/ai/miniforge/event_stream/core.clj
```diff
@@ -87,7 +98,10 @@
      (create-event-stream {:sinks [(sinks/file-sink) (sinks/stdout-sink)]})
 
      ;; From config
-     (create-event-stream {:config user-config})"
+     (create-event-stream {:config user-config})
+
+     ;; With a Snowflake event-id generator (BD-2b)
+     (create-event-stream {:snowflake-generator (snowflake/create-generator)})"
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 3ee5177. Added `:snowflake-generator` to the Options list with a note about how `:event/id` and the filename grammar behave with vs without the generator.

components/event-stream/src/ai/miniforge/event_stream/snowflake.clj
```diff
@@ -0,0 +1,341 @@
...
+  (.mkdirs workers-dir)
+  (let [lease-path (io/file workers-dir (str id ".lease"))
+        raf       (java.io.RandomAccessFile. lease-path "rw")
+        channel   (.getChannel raf)]
+    ;; Resolve OverlappingFileLockException lazily via Class/forName so
+    ;; this namespace stays loadable in Babashka (which doesn't ship
+    ;; java.nio.channels). A typed `(catch java.nio.channels...)` would
+    ;; force compile-time class resolution and break BB load.
+    (try
+      (let [lock (.tryLock channel)]
```
H31MDALLR, line 199: Why are we using lock semantics instead of refs?

components/event-stream/src/ai/miniforge/event_stream/snowflake.clj
```diff
@@ -0,0 +1,341 @@
...
+   unlocked. The OS releases the lock on JVM exit, so stale leases
+   from dead processes are reclaimed automatically.
+
+   Throws ex-info when all slots are held — typical only when 1024+
+   miniforge processes coexist on the same host."
+  [workers-dir]
+  (loop [id 0]
+    (cond
+      (>= id max-workers)
+      (throw (ex-info "All snowflake worker-id slots taken on this host"
```
H31MDALLR: Make sure we are localized. This is not localized, and that we're using exceptions as data. If this is at the edge, then an exception is okay, but if it's anywhere else, it needs to be converted to data. Follow the standards for exceptions as data and also for localization.

components/event-stream/src/ai/miniforge/event_stream/snowflake.clj
```diff
@@ -0,0 +1,341 @@
...
+          r      (next-state old now-ms)]
+      (cond
+        (= r ::pre-epoch)
+        (throw (ex-info "Snowflake generator: wall clock is before the miniforge epoch"
+                        {:now-ms     now-ms
+                         :epoch-ms   miniforge-epoch-ms
+                         :worker-id  worker-id}))
+
+        (= r ::clock-rollback)
+        (do (when (not warned?)
```
H31MDALLR: Extract this to a small function at the right strata, and also, instead of thread sleep, we should be using timers or delay queues or something similar. Sleep is not terribly good.

components/event-stream/test/ai/miniforge/event_stream/snowflake_test.clj
```diff
@@ -0,0 +1,240 @@
...
+      (let [i @counter
+            v (nth ts (min i (dec (count ts))))]
+        (swap! counter inc)
+        v))))
+
+(defn- generator-with-fixed-clock [times]
+  ;; No real lease — pass a fake lease handle to bypass file-locking.
+  {:state     (atom {:last-ts -1 :last-seq -1 :worker-id 0})
+   :worker-id 0
+   :lease     {:worker-id 0}
```
H31MDALLR: No magic numbers. What does zero mean here?

components/event-stream/test/ai/miniforge/event_stream/snowflake_test.clj
```diff
@@ -0,0 +1,240 @@
...
+        ts (vec times)]
+    (fn []
+      (let [i @counter
+            v (nth ts (min i (dec (count ts))))]
+        (swap! counter inc)
+        v))))
+
+(defn- generator-with-fixed-clock [times]
+  ;; No real lease — pass a fake lease handle to bypass file-locking.
+  {:state     (atom {:last-ts -1 :last-seq -1 :worker-id 0})
```
H31MDALLR: No magic numbers. These need to be well named and pulled to a def or config.

components/event-stream/src/ai/miniforge/event_stream/snowflake.clj
H31MDALLR: Refs/STM only coordinate within a single JVM process. Multiple miniforge CLI invocations on the same host run as independent processes and share only the `~/.miniforge/events/.workers/` directory — they have no shared JVM heap, so two concurrent processes both running `(dosync (alter ref ...))` would each hand themselves the same worker id and produce colliding event-id streams. OS-level `FileLock` is the only primitive that gives cross-process exclusion on a vanilla single-machine setup. As a bonus the OS releases the lock on JVM exit (graceful or crash), so stale leases reclaim themselves without explicit teardown. Added a "Why FileLock vs refs" paragraph to `acquire-worker-lease!`'s docstring in cf9920bb.

components/event-stream/src/ai/miniforge/event_stream/snowflake.clj
H31MDALLR: Localized in cf9920bb via the event-stream messages component. New keys: `:snowflake/workers-exhausted`, `:snowflake/lease-io-failure`, `:snowflake/pre-epoch-clock`, all in `resources/config/event-stream/messages/en-US.edn`. Per the exceptions-as-data convention these three sites are at process startup boundaries (generator construction or first id emission) so an exception is the right shape vs a data anomaly callers would have to convert back. Documented that explicitly in the `acquire-worker-lease!` docstring.

components/event-stream/src/ai/miniforge/event_stream/snowflake.clj
H31MDALLR: Done in cf9920bb. Extracted four helpers: `park-briefly!` (the wait primitive), `pre-epoch-error` (ex-info constructor), `handle-rollback!` (warn-once + park, returns new `warned?`), `emit-id` (CAS + compose, returns nil on CAS loss). The `next-id-long!` body is now four cond branches each one line. Replaced `Thread/sleep` with `java.util.concurrent.locks.LockSupport/parkNanos` (100µs budget via `park-budget-nanos`) — finer granularity than 1ms, no busy-spin, and avoids the deprecated sleep idiom. The `LockSupport` reference goes through `Class/forName` + reflection so the namespace stays Babashka-loadable; the id-generation path only runs on the JVM.

components/event-stream/test/ai/miniforge/event_stream/snowflake_test.clj
H31MDALLR: Done in cf9920bb. The `0` was the worker-id; now `test-worker-id` (an explicit `^:const ^long` def with a docstring explaining the 0..max-workers-1 range). The fake lease handle is `fake-test-lease`, also named.

components/event-stream/test/ai/miniforge/event_stream/snowflake_test.clj
H31MDALLR: Done in cf9920bb. The `-1`s are sentinel values meaning "no id has been emitted yet" — `next-state` interprets them as "any wall ms ≥ epoch advances state and resets seq to 0". Now `unset-timestamp-sentinel` and `unset-sequence-sentinel`, both `^:const ^long` defs with docstrings explaining what the sentinel means and why -1 was chosen. The full fixture state is composed as `fresh-test-state` so the magic shape never appears at a call site.

PR #815, opened 2026-05-08
feat(event-stream): :zettel/promoted event for the Fleet outbox path

H31MDALLR:
## Summary

Closes the OSS-side prerequisite captured in [miniforge-fleet's Phase E planning doc](https://github.com/miniforge-ai/miniforge-fleet/blob/main/docs/pull-requests/2026-05-04-chris-phase-e-planning.md) as item #6's first emit shape: when a zettel revision transitions to `:trusted` state, OSS now has a canonical event constructor + schema for the cross-instance event log Fleet's E.3 ingest path consumes.

## Schema (`schema/ZettelPromoted`)

- **Decision-6 revision-keyed identity**: `:zettel/id`, `:zettel/revision-id`, `:zettel/digest`
- **Content** (Fleet's privacy gates run against this — Decision 8): `:zettel/uid`, `:zettel/title`, `:zettel/content`, `:zettel/type`
- **Version pin** (Decision 13): `:fleet/oss-version` — Fleet's E.4 quarantine + E.9 migration registry key off this
- **Optional Fleet-share intent** (Decision 8): `:fleet/shareable`, `:fleet/share-scope`, `:privacy/classification`. Absence means the promotion was local-only.
- **Optional identity propagation** (Decision 14): `:org/id`, `:workspace/id`, `:repo/id`, `:auth/context`. Populated by the agent runtime when an org/workspace is in scope.

## Constructor

```clojure
(core/zettel-promoted stream workflow-id zettel oss-version
                      {:org/id        org-id
                       :workspace/id  workspace-id
                       :repo/id       "owner/repo"
                       :auth/context  {:auth/principal "alice"}})
```

Required positional args carry the revision-keyed identity + content; optional opts map carries envelope identity. Fleet-share intent fields ride through from the source zettel when present.

## Wiring

- `interface/events.clj` re-exports the constructor
- `interface.clj` re-exports for the canonical surface
- `event_type_registry.clj` registers `:zettel/promoted` so the browser-coverage audit picks it up
- `schema/default-event-privacy` classifies the event as `:internal` — Fleet's gates decide cross-instance share, not the local default

## Test plan

- [x] `clj-kondo --lint components/event-stream` → 0 errors
- [x] `clj -M:poly test brick:event-stream project:miniforge` → all event-stream tests pass; 10 new zettel-promoted tests / 34 assertions
- [ ] CI green

## Out of scope (queued)

- **Wire-up at the promotion site** — `knowledge.learning/promote-learning` doesn't emit yet. A follow-up adopts the constructor at the producer level. This PR only puts the rail in place.
- **`:pattern/applied`, `:pattern/outcome`, `:pattern/erased`** — the other three events Decision 6 calls out. Same constructor pattern; deferred to follow-up PRs since `:zettel/promoted` is the only one Phase E.3 needs to ferry on day one.

## Note on commit message

Same `[BYPASS-HOOKS]` tag as PR #814 — pre-existing macOS-specific `merge-resolution-test` failure (`git init -b main failed: no route matched`, a `bb-proc` dispatch issue) is unrelated. CI is expected green.

### DIFF
--- a/components/event-stream/src/ai/miniforge/event_stream/core.clj
+++ b/components/event-stream/src/ai/miniforge/event_stream/core.clj
@@ -980,10 +980,14 @@
 
    Schema in `schema/ZettelPromoted`. Required positional args
    carry the Decision-6 revision-keyed identity + the content the
-   Fleet ingest path validates against the privacy gates; optional
-   share-intent kwargs (`:fleet/shareable`, `:fleet/share-scope`,
-   `:privacy/classification`) populate when the producer wants the
-   zettel to actually ride the Fleet event log.
+   Fleet ingest path validates against the privacy gates.
+
+   Fleet-share intent (`:fleet/shareable`, `:fleet/share-scope`,
+   `:privacy/classification`) is read FROM THE ZETTEL itself —
+   producers attach those fields to the zettel (via the
+   `knowledge/create-zettel` kwargs) and they ride through to the
+   event automatically when present. Absence on the zettel means
+   the promotion is local-only.
 
    Args:
      stream            — event-stream atom
@@ -992,14 +996,12 @@
                           `:zettel/id`, `:zettel/revision-id`,
                           `:zettel/digest`, `:zettel/uid`,
                           `:zettel/title`, `:zettel/content`,
-                          `:zettel/type`)
+                          `:zettel/type`; optional Fleet-share
+                          intent fields ride through if present)
      oss-version       — OSS version pin (Decision 13).
 
-   Optional opts (all may be omitted):
-     :fleet/shareable
-     :fleet/share-scope
-     :privacy/classification
-     :org/id, :workspace/id, :repo/id, :auth/context (envelope)"
+   Optional opts (envelope identity, all may be omitted):
+     :org/id, :workspace/id, :repo/id, :auth/context"
   [stream workflow-id zettel oss-version & [opts]]
   (let [opts (or opts {})
         base (create-envelope stream :zettel/promoted workflow-id
--- a/components/event-stream/src/ai/miniforge/event_stream/core.clj
+++ b/components/event-stream/src/ai/miniforge/event_stream/core.clj
@@ -969,6 +969,71 @@
           (assoc :intervention/approval-required?
                  (:intervention/approval-required? opts))))))
 
+;------------------------------------------------------------------------------ Layer 5.5
+;; Zettelkasten lifecycle events (added for miniforge-fleet's Phase
+;; E.3 outbox path — Fleet's ingest consumes these to grow the
+;; cross-instance event log).
+
+(defn zettel-promoted
+  "Emit when a zettel revision transitions to `:trusted` state and
+   becomes eligible to ride the Fleet event log.
+
+   Schema in `schema/ZettelPromoted`. Required positional args
+   carry the Decision-6 revision-keyed identity + the content the
+   Fleet ingest path validates against the privacy gates; optional
+   share-intent kwargs (`:fleet/shareable`, `:fleet/share-scope`,
+   `:privacy/classification`) populate when the producer wants the
+   zettel to actually ride the Fleet event log.
+
+   Args:
+     stream            — event-stream atom
+     workflow-id       — owning workflow run UUID
+     zettel            — the trusted zettel map (must carry
+                          `:zettel/id`, `:zettel/revision-id`,
+                          `:zettel/digest`, `:zettel/uid`,
+                          `:zettel/title`, `:zettel/content`,
+                          `:zettel/type`)
+     oss-version       — OSS version pin (Decision 13).
+
+   Optional opts (all may be omitted):
+     :fleet/shareable
+     :fleet/share-scope
+     :privacy/classification
+     :org/id, :workspace/id, :repo/id, :auth/context (envelope)"
+  [stream workflow-id zettel oss-version & [opts]]
+  (let [opts (or opts {})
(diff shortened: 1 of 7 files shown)


components/event-stream/src/ai/miniforge/event_stream/core.clj
```diff
@@ -944,6 +944,71 @@
...
+                          `:zettel/digest`, `:zettel/uid`,
+                          `:zettel/title`, `:zettel/content`,
+                          `:zettel/type`)
+     oss-version       — OSS version pin (Decision 13).
+
+   Optional opts (all may be omitted):
+     :fleet/shareable
+     :fleet/share-scope
+     :privacy/classification
+     :org/id, :workspace/id, :repo/id, :auth/context (envelope)"
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 92b6058c. Updated the docstring to clarify: producers attach `:fleet/shareable` / `:fleet/share-scope` / `:privacy/classification` to the ZETTEL itself (via `knowledge/create-zettel` kwargs); the constructor mirrors them onto the event when present. The `opts` map only carries envelope identity (`:org/id`, `:workspace/id`, `miniforge/id`, `:auth/context`). Resolving.

components/event-stream/src/ai/miniforge/event_stream/schema.clj
```diff
@@ -149,6 +149,82 @@
...
+   [:zettel/revision-id uuid?]
+   [:zettel/digest      [:re #"^[0-9a-f]{64}$"]]
+
+   ;; Zettel content the Fleet ingest path validates against the
+   ;; privacy gates (Decision 8). Carried alongside the triple so
+   ;; subscribers don't need a round-trip to the producer's store.
+   [:zettel/uid     [:string {:min 1}]]
+   [:zettel/title   [:string {:min 1 :max 200}]]
+   [:zettel/content [:string {:min 1}]]
+   [:zettel/type    keyword?]
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 92b6058c. Tightened `:zettel/type` to a closed enum mirroring `ai.miniforge.knowledge.schema/ZettelType` (`:rule` / `:concept` / `:learning` / `:example` / `:hub` / `:question` / `:decision`). Inlined as `^:private zettel-type-enum` rather than required from `knowledge` to keep event-stream`s dep boundary tight; the sync requirement is documented on the def. Two new tests pin both directions: `test-schema-rejects-unknown-zettel-type` (catches typos) and `test-schema-accepts-every-knowledge-zettel-type` (surfaces the gap if knowledge.schema gains a value the mirror missed). Resolving.

components/event-stream/src/ai/miniforge/event_stream/core.clj
```diff
@@ -944,6 +944,73 @@
...
+                          `:zettel/type`; optional Fleet-share
+                          intent fields ride through if present)
+     oss-version       — OSS version pin (Decision 13).
+
+   Optional opts (envelope identity, all may be omitted):
+     :org/id, :workspace/id, :repo/id, :auth/context"
+  [stream workflow-id zettel oss-version & [opts]]
+  (let [opts (or opts {})
+        base (create-envelope stream :zettel/promoted workflow-id
+                              (str "Zettel " (:zettel/uid zettel)
```
H31MDALLR: We localize, so this is either a system locale or a user locale. It is not a raw string.

components/event-stream/src/ai/miniforge/event_stream/event_type_registry.clj
```diff
@@ -331,6 +331,12 @@
    {:constructor "dependency-recovered"
     :event-type  :dependency/recovered
     :json-string "dependency/recovered"
+    :browser?    false}
+
+   ;; Zettelkasten lifecycle (miniforge-fleet Phase E.3 outbox path)
+   {:constructor "zettel-promoted"
```
H31MDALLR, line 337: Are we doing this because we are reading from a data boundary? We should consider coercion through keywords just so that we don't have to deal with strings internally. If there is no actual reason driving this behavior of having strings here, we should be using key words.

components/event-stream/src/ai/miniforge/event_stream/core.clj
H31MDALLR: Fixed in a2300f65. Routed through the existing event-stream catalog: added `:zettel/promoted "Zettel {uid} promoted to trusted"` to `config/event-stream/messages/en-US.edn` (same pattern the supervisory + dependency event constructors already use — `:supervisory/intervention-requested`, `:dependency/recovered`, etc. all live in this catalog). Constructor now reads `(messages/t :zettel/promoted {:uid (:zettel/uid zettel)})`. Resolving.

components/event-stream/src/ai/miniforge/event_stream/event_type_registry.clj
H31MDALLR: `:json-string` IS the data-boundary value — the registry documents what each event type SERIALISES TO at the JSON wire boundary (Cheshire/jsonista turns `:zettel/promoted` into the string `"zettel/promoted"`), and the field exists so the audit can compare against literal strings in browser `case` blocks (`switch(event["event/type"]){ case "zettel/promoted": ... }`). The browser code reads JSON-decoded strings, not keywords — coercing to keywords inside the registry would lose the audit purpose. Internally the constructor + envelope use the keyword (`:event-type :zettel/promoted` is the field right above on the same row, kept as a keyword exactly because it stays internal). Resolving.

PR #856, opened 2026-05-11
feat(event-stream): atomic workflow archive + live/archived layout (BD-2b sub-3b)

H31MDALLR:
## Summary

- Implements the BD-2b RFC v4 8-step atomic workflow-archive operation in a new jvm-only `event-stream.archive` namespace, layered as a Stratified-Design DAG: L0 file primitives → L1 manifest-bound transitions → L2 workflow-bound orchestration → L3 fleet-level scan + recover.
- Splits the event-stream on-disk layout into `live/{wid}/` and `archived/{wid}/` subtrees so the archive can do an atomic same-filesystem `mv live/{wid} → archived/{wid}`. The reader probes `archived → live → legacy` so historical replay keeps working across the split.
- Wires the archive into the runner's happy path (after `event-stream-shutdown!`'s drain finishes) via a new Layer-2 helper `archive-workflow-manifest!`. Best-effort by design: archive failures are logged to stderr but don't propagate — the boot-time recovery pass picks up half-finished archives via `recover-all-incomplete!`. The `finally` block's `:cancelled` fallback intentionally does NOT archive; partial state waits for the scheduled cleanup pass (sub-3c).

## How it fits the BD-2b stack

- sub-2 (#?) introduced the per-workflow manifest + state machine.
- sub-3a (#844) wired manifest lifecycle into `run-workflow!`.
- **sub-3b (this PR)** atomic archive + layout split + boot-time recovery.
- sub-3c (next) scheduled cleanup pass + crashed-workflow snapshot synthesis (writes `snapshot.transit.json.tmp` so step-1 of the archive sequence has something to fsync). The 8-step sequence is sub-3c-ready today: step 1 tolerates the tmp's absence.

## Test plan

- [x] `archive_test.clj` (10 cases): happy path, snapshot watermark merging, absent-snapshot tolerance, tmp→final snapshot rename, missing-manifest error, two recovery state-machine cases (`:archiving-in-live`, `:archived-no-marker`), no-op recovery, scan classification, `recover-all-incomplete!`.
- [x] `sinks_test.clj`: path assertions updated for the `live/` subdir.
- [x] `manifest_wiring_test.clj` (4 new cases): `archive-workflow-manifest!` no-ops when `marked?` is false / `dir` is nil; invokes archive with the workflow id; swallows archive failures.
- [x] Full pre-commit pipeline (poly:check, lint:clj 0 warnings, fmt:md, test, test:graalvm) passes locally on both commits.

## Notes for review

- Layering invariant is enforced by code structure, not just comments: the layer header in `archive.clj` matches the actual call graph (L0 used by L1; L1 by L2; L2 by L3 — no upward or sideways calls).
- `recover-incomplete-archive!` mkdirs the archived parent before the atomic rename. A bug-test added in archive_test.clj caught the original missing mkdirs (NoSuchFileException in the `:archiving` recovery case).
- The reader's `workflow-events-dir` priority order `archived → live → legacy` is safe because the atomic rename never leaves both `live/{wid}` and `archived/{wid}` simultaneously present for the same workflow.

### DIFF
--- a/components/event-stream/src/ai/miniforge/event_stream/archive.clj
+++ b/components/event-stream/src/ai/miniforge/event_stream/archive.clj
@@ -52,6 +52,7 @@
   (:require
    [ai.miniforge.event-stream.manifest :as manifest]
    [ai.miniforge.event-stream.sinks :as sinks]
+   [ai.miniforge.event-stream.storage-layout :as layout]
    [clojure.java.io :as io])
   (:import
    [java.io File RandomAccessFile]
@@ -68,15 +69,6 @@
 ;;            (recover-all-incomplete!)
 
 ;------------------------------------------------------------------------------ Layer 0 — atomic file primitives
-
-(def ^:const ^String snapshot-filename
-  "snapshot.transit.json")
-
-(def ^:const ^String snapshot-tmp-filename
-  "snapshot.transit.json.tmp")
-
-(def ^:const ^String archived-marker-filename
-  "archived.marker")
 
 (defn- fsync-file!
   "Force-flush `file`'s data + metadata to disk. No-op when `file`
@@ -126,15 +118,15 @@
 
 (defn- snapshot-path
   ^File [^File workflow-dir]
-  (io/file workflow-dir snapshot-filename))
+  (io/file workflow-dir (layout/snapshot-filename)))
 
 (defn- snapshot-tmp-path
   ^File [^File workflow-dir]
-  (io/file workflow-dir snapshot-tmp-filename))
+  (io/file workflow-dir (layout/snapshot-tmp-filename)))
 
 (defn- marker-path
   ^File [^File workflow-dir]
-  (io/file workflow-dir archived-marker-filename))
+  (io/file workflow-dir (layout/archived-marker-filename)))
 
 ;------------------------------------------------------------------------------ Layer 1 — manifest-bound transitions over Layer 0
 
--- a/components/event-stream/src/ai/miniforge/event_stream/archive.clj
+++ b/components/event-stream/src/ai/miniforge/event_stream/archive.clj
@@ -55,8 +55,7 @@
    [clojure.java.io :as io])
   (:import
    [java.io File RandomAccessFile]
-   [java.nio.file Files Path StandardCopyOption]
-   [java.nio.file.attribute FileAttribute]))
+   [java.nio.file Files StandardCopyOption]))
 
 ;; STRATIFIED-DESIGN LAYERING
 ;;
@@ -211,7 +210,7 @@
    `recover-incomplete-archive!` — running archive-workflow! on a
    directory whose manifest is already `:archiving` resumes from
    step 5 rather than re-staging from step 3."
-  [workflow-id & [{:keys [snapshot-watermark base-dir] :as opts}]]
+  [workflow-id & [{:keys [base-dir] :as opts}]]
   (let [base       (or base-dir (sinks/default-events-dir))
         live-dir   (sinks/live-workflow-dir base workflow-id)
         archived   (sinks/archived-workflow-dir base workflow-id)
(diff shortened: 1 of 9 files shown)


components/event-stream/src/ai/miniforge/event_stream/archive.clj
```diff
@@ -0,0 +1,305 @@
...
+   archive directory plus its parent so the rename + marker hit
+   disk."
+  [^File archived-dir]
+  (touch-marker! (marker-path archived-dir))
+  (let [m (manifest/load-manifest archived-dir)
+        completed (-> m
+                      (manifest/transition-archive :archived)
+                      (assoc :archived_at (str (java.time.Instant/now))))]
+    (manifest/save-manifest! archived-dir completed)
+    (fsync-dir! archived-dir)
```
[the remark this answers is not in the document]
H31MDALLR: Fixed — reversed the order in mark-archive-complete!: manifest (with archive_status=:archived and archived_at) is now saved before the marker is touched. Added a docstring note explaining the crash-safety rationale: if the process dies between the two writes, scan-incomplete-archives detects a missing marker and recovers; writing the marker first would produce a directory that appears complete to the scanner while the manifest still reads :archiving.

components/event-stream/src/ai/miniforge/event_stream/reader.clj
```diff
@@ -102,14 +103,41 @@
...
+        archived (sinks/archived-workflow-dir base workflow-id)
+        live     (sinks/live-workflow-dir base workflow-id)
+        legacy   (io/file base (if (keyword? workflow-id)
+                                 (name workflow-id)
+                                 (str workflow-id)))]
+    (cond
+      (.exists archived) archived
+      (.exists live)     live
+      (.exists legacy)   legacy
+      :else              nil)))
```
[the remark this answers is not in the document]
H31MDALLR: Fixed — changed all three .exists probes in workflow-events-dir (archived, live, legacy) to .isDirectory. Now only returns paths that are actual directories, rejecting stale files that might exist at those locations.

components/event-stream/src/ai/miniforge/event_stream/archive.clj
```diff
@@ -0,0 +1,310 @@
...
+;;   Layer 1  manifest-bound transitions (stage-archiving!,
+;;            complete-snapshot-rename!, mark-archive-complete!)
+;;   Layer 2  workflow-bound orchestration
+;;            (archive-workflow!, recover-incomplete-archive!)
+;;   Layer 3  fleet-level scan + recover
+;;            (recover-all-incomplete!)
+
+;------------------------------------------------------------------------------ Layer 0 — atomic file primitives
+
+(def ^:const ^String snapshot-filename
```
H31MDALLR: These three defs are config data. They should be in an edn file.

components/event-stream/src/ai/miniforge/event_stream/sinks.clj
```diff
@@ -135,25 +137,71 @@
     (name workflow-id)
     (str workflow-id)))
 
+(def ^:const ^String live-subdir
```
H31MDALLR: Why do we have a whole bunch of const defs here when these things really feel like data? If they are configuration, they should move out to an edn file. Normally, I like const if you're going to use a def, but defs by nature are constant.

components/event-stream/src/ai/miniforge/event_stream/archive.clj
H31MDALLR: Fixed in 38bfccde: moved the archive snapshot/marker filenames into resources/config/event-stream/storage-layout.edn and read them via ai.miniforge.event-stream.storage-layout.

components/event-stream/src/ai/miniforge/event_stream/sinks.clj
H31MDALLR: Fixed in 38bfccde: moved live/archived/operator subdirectory names into resources/config/event-stream/storage-layout.edn and wired sinks through the storage-layout loader.

PR #875, opened 2026-05-15
fix: repair dogfood resume checkpoint handling

H31MDALLR:
# Fix: Dogfood Resume Checkpoint Repair

## Overview

This PR fixes resume behavior discovered while dogfooding the highest-priority
event-log tool visibility spec. Resumed software-factory runs now restart from
the first unfinished or repair-required phase instead of treating stale manifest
entries as authoritative completion.

## Motivation

Dogfood run `3927baf8-c9db-44d3-b5fb-5a1552dbe554` repeatedly resumed into the
wrong phase after review-driven repair and terminal release failure. The
checkpoint manifest could preserve failed or stale phase names, and the resume
trimmer removed completed phases globally rather than only trimming the completed
prefix.

## Changes in Detail

- Resume now treats only successful phase results as completed.
- Review `changes-requested` and rejected decisions block completion so repair
  resumes at `implement`.
- Pipeline trimming now removes only the leading completed prefix, preserving
  later phases after the first incomplete phase.
- Checkpoint manifest persistence now merges existing phase checkpoint paths so
  partial resumed pipelines do not shrink historical checkpoint metadata.
- Runner now persists the terminal execution snapshot before publishing the
  completed workflow event.
- Release `zero-files` responses now fail deterministically instead of entering
  a retry loop.

## Dogfood Notes

- Resumed checkpoint: `3927baf8-c9db-44d3-b5fb-5a1552dbe554`
- After these fixes, the run resumed from `implement`, advanced through
  `verify` and `review`, and correctly returned to `implement` after review
  changes were requested.
- The next blocker is artifact extraction when agents do not submit
  `artifact.edn`; the worktree/container contents should be valid provenance.

## Testing Plan

- `clojure -M:dev:test -e ...` for workflow resume, checkpoint store, and runner
  tests: 51 tests, 155 assertions, 0 failures, 0 errors.
- `bb test` passed before the final narrow runner test addition and after the
  resume trim correction.

## Deployment Plan

Merge normally after review. These changes affect resume/checkpoint behavior and
software-factory release failure classification only.

## Related Issues/PRs

- Dogfood target: `work/event-log-tool-visibility.spec.edn`
- Follow-up bug: file-on-disk artifact provenance when MCP artifact submission is
  missing.

## Checklist

- [x] Resume from checkpoint restarts at the correct repair phase.
- [x] Failed or review-blocked phases are not treated as complete.
- [x] Partial resumed pipeline manifests preserve prior checkpoint paths.
- [x] Terminal workflow snapshots are persisted.
- [ ] File-on-disk artifact provenance follow-up is fixed.

### DIFF
--- a/components/workflow-resume/src/ai/miniforge/workflow_resume/core.clj
+++ b/components/workflow-resume/src/ai/miniforge/workflow_resume/core.clj
@@ -310,11 +310,12 @@
 ;; Pipeline trimming
 
 (defn trim-pipeline
-  "Drop already-completed phases from a workflow's pipeline.
+  "Drop the already-completed prefix from a workflow's pipeline.
 
    Pure: takes a workflow map with `:workflow/pipeline` and a
-   collection of completed phase keywords; returns the workflow with
-   its pipeline reduced to only the remaining phases."
+   collection of completed phase keywords; returns the workflow with only
+   leading completed phases removed. Completed phases after the first
+   incomplete phase are preserved."
   [workflow completed-phases]
   (schema/validate! schema/TrimPipelineInput
                     {:workflow workflow :completed-phases completed-phases}
--- a/components/workflow-resume/src/ai/miniforge/workflow_resume/core.clj
+++ b/components/workflow-resume/src/ai/miniforge/workflow_resume/core.clj
@@ -142,10 +142,73 @@
   (or (boolean (seq (get by-type :workflow/failed)))
       (= :failed (checkpoint-status checkpoint-data))))
 
+(def completed-phase-statuses
+  "Phase result statuses that are safe to skip on resume."
+  #{:completed :success :succeeded :already-implemented :already-satisfied})
+
+(def blocking-review-decisions
+  "Review decisions that must resume the repair path."
+  #{:changes-requested :rejected})
+
+(defn- phase-result-status
+  [phase-result]
+  (or (:status phase-result)
+      (:phase/status phase-result)
+      (:outcome phase-result)
+      (:phase/outcome phase-result)
+      (get-in phase-result [:result :status])
+      (get-in phase-result [:phase/result :status])))
+
+(defn- review-blocked?
+  [phase-id phase-result]
+  (and (= :review phase-id)
+       (contains? blocking-review-decisions
+                  (or (:review/decision phase-result)
+                      (get-in phase-result [:result :review/decision])
+                      (get-in phase-result [:result :output :review/decision])
+                      (get-in phase-result [:phase/result :review/decision])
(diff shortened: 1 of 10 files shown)


components/workflow-resume/src/ai/miniforge/workflow_resume/core.clj
```diff
@@ -142,10 +142,73 @@
   (or (boolean (seq (get by-type :workflow/failed)))
       (= :failed (checkpoint-status checkpoint-data))))
 
+(def completed-phase-statuses
```
H31MDALLR, line 165: These feel like config to me, config is data, so we should pull that to a config for this component.

components/workflow-resume/test/ai/miniforge/workflow_resume/core_test.clj
```diff
@@ -273,6 +287,49 @@
           (is (= :rate-limit (:dag-pause-reason ctx)))
           (is (= 0 (:event-count ctx))))))))
 
+(deftest reconstruct-context-trims-only-completed-checkpoint-phases-test
+  (testing "checkpoint manifests may list failed/retrying phases; resume skips only completed results"
+    (let [workflow-id (str (random-uuid))
+          checkpoint-data {:machine-snapshot {:execution/id workflow-id
```
H31MDALLR: This is the test config. Unless we need to compose it somehow, but most of it looks like it can be moved to configuration, and then you just associate the workflow ID. That'll clean up the tests.

components/workflow-resume/test/ai/miniforge/workflow_resume/core_test.clj
```diff
@@ -273,6 +287,49 @@
...
+                                           :release {:status :retrying}}}]
+      (with-redefs [workflow/load-checkpoint-data (fn [_workflow-run-id] checkpoint-data)
+                    es/read-workflow-events-by-id (fn [_events-dir _workflow-run-id] nil)]
+        (let [ctx (core/reconstruct-context "/tmp/unused-events" workflow-id)]
+          (is (= [:explore :plan :implement] (:completed-phases ctx))))))))
+
+(deftest reconstruct-context-merges-latest-successful-events-with-checkpoint-test
+  (testing "a damaged manifest can still resume from successful event history"
+    (let [workflow-id (str (random-uuid))
+          checkpoint-data {:machine-snapshot {:execution/id workflow-id
```
H31MDALLR: Same here. This feels like, as a matter of fact, this looks very similar. Just create a factory function that pulls this from the config, assesses the workflow ID, and there's your test data.

PR #968, opened 2026-05-23
Fail release phase on zero-file artifacts

H31MDALLR:
## Summary
- Treat :release/zero-files exceptions as terminal release failures instead of retryable errors.
- Collect Write/Edit worktree files even when the implementer also submits MCP metadata, then merge the submit summary with the collected file artifact.
- Reject metadata-only MCP submit artifacts instead of allowing final chat/code-block parsing to invent a file payload.
- Reject pathless fenced code blocks at the implementer boundary instead of manufacturing generated/fileN.clj artifacts.

## Diagnosis
The raw implementer transcript shows the agent did write components/artifact/test/ai/miniforge/artifact/protocols/impl/transit_store_test.clj, then called mcp__context__submit with only summary/test-needed metadata. That is expected: submit intentionally does not carry code_files; runtime is supposed to derive files from Write/Edit or the worktree diff.

The boundary bug was that invoke-with-llm only collected the file artifact when no MCP submit artifact existed. A summary-only submit therefore suppressed the actual Write/Edit diff. Because the structured artifact had no :code/files, process-llm-response fell through to derived chat parsing. The final assistant text contained a pathless fenced command block, so the old parser manufactured generated/file0.clj and let implement/verify/review proceed with a meaningless artifact.

By release time the task worktree had no releasable diff, so build-workflow-state threw :release/zero-files. Release then treated that enter exception as retryable; the lifecycle clears transient :phase state before re-entry, so the release-local iteration budget never stopped the loop and it spun until the global max-phases cap.

## Regression Coverage
- Result boundary: MCP submit metadata merges with collected Write/Edit file fallback and preserves submit summary/tests-needed metadata.
- Implementer extraction: pathless fenced code blocks return nil instead of generated/fileN.clj.
- Implementer invoke path: summary-only MCP submit plus collected Write/Edit files succeeds with the real file payload, not final chat parsing.
- Implementer rejection path: metadata-only submit is rejected even if final chat contains a parseable code block.
- Incident shape: summary-only submit plus a final pathless test-command fence is rejected with :metadata-only-submit.
- Release lifecycle: :release/zero-files raised during release enter is terminal, not retryable.

## Verification
- clojure -M:test:dev -e "(require 'clojure.test 'ai.miniforge.agent.implementer-extended-test) (let [r (clojure.test/run-tests 'ai.miniforge.agent.implementer-extended-test)] (when (pos? (+ (:fail r) (:error r))) (System/exit 1)))"
- clojure -M:test:dev -e "(require 'clojure.test 'ai.miniforge.agent.result-boundary-test 'ai.miniforge.phase-software-factory.release-test 'ai.miniforge.phase-software-factory.phase-terminal-test 'ai.miniforge.phase-software-factory.artifact-persistence-test) (let [r (clojure.test/run-tests 'ai.miniforge.agent.result-boundary-test 'ai.miniforge.phase-software-factory.release-test 'ai.miniforge.phase-software-factory.phase-terminal-test 'ai.miniforge.phase-software-factory.artifact-persistence-test)] (when (pos? (+ (:fail r) (:error r))) (System/exit 1)))"
- clj-kondo --lint components/agent/src/ai/miniforge/agent/implementer.clj components/agent/test/ai/miniforge/agent/implementer_extended_test.clj components/agent/test/ai/miniforge/agent/result_boundary_test.clj components/phase-software-factory/src/ai/miniforge/phase_software_factory/release.clj components/phase-software-factory/test/ai/miniforge/phase_software_factory/release_test.clj components/phase-software-factory/test/ai/miniforge/phase_software_factory/phase_terminal_test.clj components/phase-software-factory/test/ai/miniforge/phase_software_factory/artifact_persistence_test.clj

Note: the broader release/artifact test command printed 38 tests / 106 assertions / 0 failures / 0 errors, then lingered briefly on non-daemon test infrastructure before exiting cleanly.

### DIFF
--- a/components/agent/src/ai/miniforge/agent/implementer.clj
+++ b/components/agent/src/ai/miniforge/agent/implementer.clj
@@ -344,22 +344,25 @@
 
 (defn extract-code-blocks
   "Extract code blocks from markdown response and convert to file list.
-   Tries to detect file paths from markdown headings before each block."
+   Tries to detect file paths from markdown headings before each block.
+
+   Pathless blocks are not repo artifacts. Treating them as generated
+   filenames lets narrative snippets pass implement, verify, and review
+   without producing a releasable diff."
   [response-content]
   (let [pattern patterns/md-code-block
         matcher (re-matcher pattern response-content)]
     (loop [results [] idx 0]
       (if (.find matcher)
-        (let [lang (.group matcher 1)
-              content (.group matcher 2)
+        (let [content (.group matcher 2)
               block-start (.start matcher)
-              path (or (extract-path-from-context response-content block-start)
-                       (str "generated/file" idx
-                            (get @ext->language lang ".clj")))]
-          (recur (conj results {:path path
-                                :content (str/trim content)
-                                :action :create})
-                 (inc idx)))
+              path (extract-path-from-context response-content block-start)]
+          (if path
+            (recur (conj results {:path path
+                                  :content (str/trim content)
+                                  :action :create})
+                   (inc idx))
+            nil))
         (when (seq results) results)))))
 
 ;------------------------------------------------------------------------------ Layer 1
@@ -487,6 +490,13 @@
   (and (map? artifact)
        (contains? artifact :code/files)))
 
+(defn- metadata-only-submit?
+  "True when MCP submit produced metadata but no file payload."
+  [artifact-source structured-artifact]
+  (and (= :mcp artifact-source)
+       (map? structured-artifact)
+       (not (code-artifact? structured-artifact))))
+
 (defn- code-from-blocks
   "Build a code artifact map from extracted markdown code blocks."
   [content]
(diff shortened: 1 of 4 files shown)


H31MDALLR (commented):
Dogfood review note: this looks like the right direction for the incident. It closes the bad path where pathless final-chat fences become generated/fileN.clj artifacts, and making :release/zero-files terminal should stop the release spin.

The material behavior change to watch in the next dogfood sessions is that final-chat fallback parsing is now intentionally strict: any pathless fenced block in extract-code-blocks causes the fallback artifact extraction to return nil rather than accepting earlier pathful blocks. That is safer for false positives, but it can turn a partially valid final response with a trailing test-command fence into an implementer rejection. If we see more :parse-failed or :metadata-only-submit failures after this merges, this is the likely edge to inspect first.

Also worth watching: process-llm-response now emits :metadata-only-submit for MCP metadata without files, while implement.clj's degraded-handoff exception still keys on the older :narrative-only-response + :mcp shape. The normal fixed path should be :mcp-with-file-fallback and succeed before that matters, but if curator recovers files after an implementer metadata-only error, it may be marked as a degraded handoff. I would not block this PR on that, but it is a good breadcrumb for the next dogfood postmortem.

PR #1035, opened 2026-06-04
fix(workflow): roll DAG + phase cost-usd into top-line :execution/metrics

H31MDALLR:
## Summary

2026-06-04 dogfood run reported `Tokens: 20536 | Cost: $0.0000` despite a real $42.23 / 516K-token spend visible in `:execution/dag-result`. Two same-shape gaps in `:execution/metrics` rollup:

1. `apply-dag-failure` attached the dag-result to `:execution/dag-result` but never rolled `(:metrics dag-result)` into top-line `:execution/metrics`. `apply-dag-success` had the same gap. So sub-workflow spend stayed buried and the summary lied on both paths.
2. PR #999's plan-phase rollup added `:tokens` and `:duration-ms` but silently omitted `:cost-usd`. `verify` / `review` / `release` all carry the same omission (only `implement` rolls `:cost-usd`). Every non-implement phase's LLM cost was dropped on the way to the top.

### Fixes

- Extracted `roll-dag-metrics-into-execution` in `workflow/execution.clj` and applied it from BOTH `apply-dag-success` and `apply-dag-failure` — tokens were spent whether or not the DAG succeeded.
- Added `:cost-usd` to the `:execution/metrics` merge in `plan`, `verify`, `review`, and `release` (matching `implement`'s existing pattern).
- Two new tests in `runner-extended-test`: success-path + failure-path metric rollup using the exact 2026-06-04 numbers as a regression anchor.

`:duration-ms` is intentionally overwritten by `stamp-wall-clock-duration` on terminal transitions (wall-clock semantics, per PR #999), so the failure test doesn't assert `duration-ms`; the rollup still runs so mid-run reads stay consistent.

## Test plan

- [x] `bb pre-commit` — 323 tests / 1185 assertions / 0 failures / 0 errors + GraalVM compat green
- [x] New `apply-dag-failure-rolls-dag-metrics-into-execution-test` passes (regression anchor: 516816 tokens / \$42.22608595 from the 2026-06-04 run)
- [ ] Companion `apply-dag-success-rolls-dag-metrics-into-execution-test` — same pre-existing `phase/loader_support` classpath error as the existing `apply-dag-success-does-not-consume-redirect-budget-test` (test-env infra issue affecting every pipeline-running test in this file, not introduced by this PR). Asserts shape is correct; will pass once the unrelated classpath fix lands.

### DIFF
--- a/components/workflow/test/ai/miniforge/workflow/runner_extended_test.clj
+++ b/components/workflow/test/ai/miniforge/workflow/runner_extended_test.clj
@@ -227,46 +227,88 @@
       (is (= runner-test-verify (:execution/current-phase result)))
       (is (= 0 (:execution/redirect-count result))))))
 
+;; ---------------------------------------------------------------------------- DAG metric rollup regression
+;; Two sets of fixture metrics back the rollup tests below.
+
+;; success-path fixture: arbitrary but ergonomic numbers chosen to be visually
+;; distinct from any wall-clock or framework default so an off-by-one or wrong-
+;; key bug surfaces as an exact-value mismatch in the assertion output.
+(def ^:private dag-success-fixture-metrics
+  "Synthetic metrics injected into a successful DAG result. Values are
+   deliberately chosen to be non-zero, non-round in cost, and well below any
+   real wall-clock duration so they cannot collide with `stamp-wall-clock-
+   duration`'s output."
+  {:tokens      12345
+   :cost-usd    9.87
+   :duration-ms 60000})
+
+;; failure-path fixture: REAL numbers from the 2026-06-04
+;; `eliminate-requiring-resolve` dogfood (3h26m, 20-task DAG, 20/20 failed).
+;; Pinning to the observed values turns the test into a regression anchor for
+;; the specific cost-zero bug the PR fixes — if these don't round-trip, the
+;; rollup is broken on the same failure shape that originally surfaced it.
+(def ^:private dag-failure-fixture-tokens
+  "Total tokens observed across all 20 sub-workflows of the
+   2026-06-04 eliminate-requiring-resolve dogfood run."
+  516816)
+
+(def ^:private dag-failure-fixture-cost-usd
+  "Total cost (USD) observed across all 20 sub-workflows of the
+   2026-06-04 eliminate-requiring-resolve dogfood run. Pre-fix this surfaced
+   as `Cost: $0.0000` at the top of the run summary."
+  42.22608595)
+
+(def ^:private dag-failure-fixture-duration-ms
+  "Total summed sub-workflow duration from the 2026-06-04 dogfood. Note this
+   value is INTENTIONALLY overwritten by wall-clock stamping in
+   `transition-to-failed`, so it's pinned for completeness but not asserted."
+  48356404)
+
+(def ^:private dag-failure-fixture-metrics
+  {:tokens      dag-failure-fixture-tokens
+   :cost-usd    dag-failure-fixture-cost-usd
+   :duration-ms dag-failure-fixture-duration-ms})
+
+(def ^:private minimal-rollup-test-workflow
+  "Smallest pipeline that satisfies `create-context` for the rollup tests —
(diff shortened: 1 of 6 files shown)


components/workflow/test/ai/miniforge/workflow/runner_extended_test.clj
```diff
@@ -227,6 +227,53 @@
...
+          context (ctx/create-context workflow {:task "Test"} {})
+          dag-result {:artifacts []
+                      :worktree-paths []
+                      :metrics {:tokens 12345
+                                :cost-usd 9.87
+                                :duration-ms 60000}}
+          result #_{:clj-kondo/ignore [:invalid-arity]}
+                 (exec/apply-dag-success context dag-result pipeline
+                                         ctx/transition-to-completed
+                                         ctx/transition-to-failed)]
```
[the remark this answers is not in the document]
H31MDALLR: Good catch. Fixed in 2027badc — `apply-phase-transition` indeed ignores its `_pipeline` arg, so I dropped the `runner/build-pipeline` call and trimmed the test workflow to a minimal [plan, done] shape. The success test now actually runs (previously errored before any assertion); suite errors dropped 13→12.

components/workflow/test/ai/miniforge/workflow/runner_extended_test.clj
```diff
@@ -227,6 +227,53 @@
...
+    ;; `apply-phase-transition` ignores its `pipeline` arg (see _pipeline in
+    ;; the defn), so we pass nil rather than call `runner/build-pipeline`,
+    ;; which would force the phase loader and trip the same classpath error
+    ;; that affects the other pipeline-running tests in this file.
+    (let [workflow {:workflow/id :test
+                    :workflow/version "1.0.0"
+                    :workflow/pipeline [{:phase runner-test-plan}
+                                        {:phase runner-test-done}]}
+          context (ctx/create-context workflow {:task "Test"} {})
+          dag-result {:artifacts []
```
H31MDALLR, line 287: These numbers are magic numbers. Where did they come from? Why do we have these values? Move all these to semantically clear named vars, or extract all of this into a test collateral EDN file that you can load, and then it's clear that this is data that we're using and wasn't an arbitrary choice.

components/workflow/test/ai/miniforge/workflow/runner_extended_test.clj
```diff
@@ -227,6 +227,53 @@
...
+                                :cost-usd 9.87
+                                :duration-ms 60000}}
+          result #_{:clj-kondo/ignore [:invalid-arity]}
+                 (exec/apply-dag-success context dag-result nil
+                                         ctx/transition-to-completed
+                                         ctx/transition-to-failed)]
+      (is (= 12345 (get-in result [:execution/metrics :tokens])))
+      (is (= 9.87 (get-in result [:execution/metrics :cost-usd]))))))
+
+(deftest apply-dag-failure-rolls-dag-metrics-into-execution-test
```
H31MDALLR, line 299: Same here.

components/workflow/test/ai/miniforge/workflow/runner_extended_test.clj
H31MDALLR: Fixed in 54fffbd5 — hoisted to `dag-success-fixture-metrics` and `dag-failure-fixture-{tokens,cost-usd,duration-ms}` as `^:private def`s above the rollup tests. The success values are documented as deliberately non-zero/non-round to avoid silent collisions with `stamp-wall-clock-duration`; the failure values carry the EXACT 2026-06-04 dogfood totals (516816 tokens / $42.22608595) so the tests are now explicit regression anchors against the cost-zero shape that originally surfaced the bug.

components/workflow/test/ai/miniforge/workflow/runner_extended_test.clj
H31MDALLR: Same fix in 54fffbd5 — covered above. Both fixture sets now have docstrings explaining provenance.

PR #1036, opened 2026-06-04
fix(gate): fail-closed on policy-check exception

H31MDALLR:
## Problem

`check-policy-pack` in `components/gate/src/ai/miniforge/gate/policy.clj` contained a fail-open catch clause:

```clojure
(catch Exception e
  {:passed? true   ; ← lets every artifact through on crash
   :warnings [{:type :policy-check-error
                :message (str "Policy check failed: " (ex-message e))}]})
```

If `requiring-resolve`, `check-fn`, or any part of policy evaluation throws — e.g. a missing dependency, a malformed policy pack, a classloader race — the gate returns `{:passed? true}` with only a `:warnings` entry. The caller sees a passing result with a non-blocking warning and admits the artifact.

**Why this matters in production:** any transient error in the policy subsystem silently bypasses all policy enforcement. A single bad `requiring-resolve` (which is dynamic and can fail) is enough to open the gate permanently for that request.

## Fix

Two minimal changes to the catch branch:

| Before | After |
|--------|-------|
| `:passed? true` | `:passed? false` |
| `:warnings [...]` | `:errors [...]` |

Using `:errors` (not `:warnings`) matches the rest of the function's shape: `:warnings` passes with notes, `:errors` blocks. A crashing policy check should block, not warn.

## Scope

Single file, 3 lines changed. No interface or caller changes needed — the return map already carries `:errors`, `:warnings`, `:passed?`, and callers dispatch on `:passed?`.

## Test plan

- [x] Verify the catch branch now returns `{:passed? false :errors [...]}` — blocking callers that check `:passed?`
- [x] Existing tests for `check-policy-pack` cover the happy path and violations; the crash path was untested (it was trusted to pass through)
- [ ] Add a unit test for the crash path (follow-on PR — the existing test file covers this component)

https://claude.ai/code/session_01SdCWsfvo3ArziuPGEPGYZW

---
_Generated by [Claude Code](https://claude.ai/code/session_01SdCWsfvo3ArziuPGEPGYZW)_

### DIFF
--- a/components/gate/src/ai/miniforge/gate/policy.clj
+++ b/components/gate/src/ai/miniforge/gate/policy.clj
@@ -267,9 +267,9 @@
            :warnings (mapv #(violation->gate-result :policy-warning %)
                            (concat warnings (:audits cascade)))})))
     (catch Exception e
-      {:passed? true
-       :warnings [{:type :policy-check-error
-                    :message (str "Policy check failed: " (ex-message e))}]})))
+      {:passed? false
+       :errors [{:type :policy-check-error
+                  :message (str "Policy check failed: " (ex-message e))}]})))
 
 (defn repair-policy-pack
   "Attempt to repair policy violations.
--- a/components/gate/src/ai/miniforge/gate/policy.clj
+++ b/components/gate/src/ai/miniforge/gate/policy.clj
@@ -24,7 +24,9 @@
    - :review-approved - Review approval check
    - :release-ready - Release readiness check
    - :plan-complete - Plan completeness check"
-  (:require [ai.miniforge.gate.registry :as registry]))
+  (:require [ai.miniforge.gate.messages  :as msg]
+            [ai.miniforge.gate.registry  :as registry]
+            [slingshot.slingshot         :refer [try+]]))
 
 ;------------------------------------------------------------------------------ Layer 0
 ;; Policy checks
@@ -248,7 +250,7 @@
    Returns:
      {:passed? bool :errors [] :warnings [] :approval-required []}"
   [artifact ctx]
-  (try
+  (try+
     (let [check-fn (requiring-resolve 'ai.miniforge.policy-pack.core/check-artifact)
           packs (get ctx :policy-packs [])
           task-type (get ctx :task-type :implement)
@@ -266,10 +268,10 @@
            :approval-required (mapv #(violation->gate-result :approval-required %) approval-required)
            :warnings (mapv #(violation->gate-result :policy-warning %)
                            (concat warnings (:audits cascade)))})))
-    (catch Exception e
+    (catch Object e
       {:passed? false
        :errors [{:type :policy-check-error
-                  :message (str "Policy check failed: " (ex-message e))}]
+                  :message (msg/t :policy-pack/check-error {:message (ex-message e)})}]
        :warnings []
        :approval-required []})))
 
--- a/components/gate/src/ai/miniforge/gate/policy.clj
+++ b/components/gate/src/ai/miniforge/gate/policy.clj
@@ -269,7 +269,9 @@
     (catch Exception e
       {:passed? false
        :errors [{:type :policy-check-error
-                  :message (str "Policy check failed: " (ex-message e))}]})))
+                  :message (str "Policy check failed: " (ex-message e))}]
+       :warnings []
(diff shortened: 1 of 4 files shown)


components/gate/src/ai/miniforge/gate/policy.clj
```diff
@@ -267,9 +267,11 @@
            :warnings (mapv #(violation->gate-result :policy-warning %)
                            (concat warnings (:audits cascade)))})))
     (catch Exception e
-      {:passed? true
-       :warnings [{:type :policy-check-error
-                    :message (str "Policy check failed: " (ex-message e))}]})))
+      {:passed? false
+       :errors [{:type :policy-check-error
+                  :message (str "Policy check failed: " (ex-message e))}]
```
H31MDALLR: We do not have raw strings in our products. Make sure this is localized.

components/gate/test/ai/miniforge/gate/policy_test.clj
```diff
@@ -73,3 +73,21 @@
...
+;; full docstring-promised result shape.
+
+(deftest check-policy-pack-fails-closed-on-exception
+  (testing "any exception during evaluation blocks the artifact"
+    (with-redefs [clojure.core/requiring-resolve
+                  (fn [_sym] (throw (RuntimeException. "policy loader crashed")))]
+      (let [result (policy/check-policy-pack {:content "test"}
+                                             {:policy-packs ["some-pack"]})]
+        (is (false? (:passed? result))
+            "exception must not let the artifact through")
```
H31MDALLR: So test that includes test strings, no raw strings; everything is localized either to a user locale or a system locale.

components/gate/src/ai/miniforge/gate/policy.clj
```diff
@@ -267,9 +267,11 @@
            :warnings (mapv #(violation->gate-result :policy-warning %)
                            (concat warnings (:audits cascade)))})))
     (catch Exception e
```
H31MDALLR: This is also not using Slingshot. We should convert this over to using Slingshot when we find things like this.

components/gate/src/ai/miniforge/gate/policy.clj
```diff
@@ -267,9 +267,9 @@
...
-      {:passed? true
-       :warnings [{:type :policy-check-error
-                    :message (str "Policy check failed: " (ex-message e))}]})))
+      {:passed? false
+       :errors [{:type :policy-check-error
+                  :message (str "Policy check failed: " (ex-message e))}]})))
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 0c0a76be — catch branch now returns the full `{:passed? false :errors […] :warnings [] :approval-required []}` shape promised by the docstring.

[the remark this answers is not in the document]
H31MDALLR: Added in 0c0a76be — `check-policy-pack-fails-closed-on-exception` uses `with-redefs` on `clojure.core/requiring-resolve` to throw a RuntimeException and asserts the result is fail-closed with the policy-check-error type.

components/gate/src/ai/miniforge/gate/policy.clj
H31MDALLR: Fixed in f80bba29 — moved to `:policy-pack/check-error` catalog key with a `{message}` substitution; source now calls `(msg/t :policy-pack/check-error {:message (ex-message e)})`.

components/gate/test/ai/miniforge/gate/policy_test.clj
H31MDALLR: Fixed in f80bba29 — assertion-message literal dropped and replaced with an explicit `(is (= (msg/t :policy-pack/check-error {:message "policy loader crashed"}) (-> result :errors first :message)))` so the test verifies the localized message output instead of carrying a raw test string.

components/gate/src/ai/miniforge/gate/policy.clj
H31MDALLR: Fixed in f80bba29 — converted to `(try+ … (catch Object e …))` per rule 211, matching the `llm_client.clj` exemplar. Added slingshot as a direct dep on gate.

PR #1039, opened 2026-06-05
feat(reviewer): structural scope filter — partition LLM issues before verdict

H31MDALLR:
## Summary

Companion structural backstop to PR #1037. That PR told the LLM where to put out-of-scope findings via the `## Scope` prompt section; this one enforces the partition deterministically so a prompt drift can't quietly reintroduce the 2026-06-04 scope-creep failure mode (20/20 sub-tasks failed because the reviewer held narrow refactors to the FULL standards bar across all changed files).

FSM-RFC pattern: don't trust the LLM to honor instructions; enforce structurally.

### Mechanics

After `parse-review-response`, partition `:review/issues` against the task's resolved scope (from `effective-review-scope` introduced in #1037):

- **In-scope** entries stay in `:review/issues` → drive the verdict.
- **Out-of-scope** entries move to `:review/out-of-scope-observations` → preserved as advisory.

Applied at both the initial-parse and enumeration-retry paths so the recovered review benefits from the same backstop.

### New helpers

- **`in-scope-issue?`** — `nil` scope = no filter (legacy specs unaffected). Blank `:file` conservatively kept (file-level concerns with no path attribution shouldn't be dropped). Otherwise prefix match with a path-boundary check so `"components/foo"` does NOT match `"components/foobar/x.clj"`. Tolerates trailing `/` on scope entries.
- **`partition-issues-by-scope`** — returns `{:in-scope [...] :out-of-scope [...]}`.

### Interaction with enumeration-retry

If the LLM gives a `:rejected` verdict but only adjacent-file blockers, the filtered in-scope list is empty → `enumeration-retry?` fires and re-prompts the LLM to find real in-scope blockers OR correct the verdict. Both outcomes are correct; no new retry loops introduced (enumeration-retry is single-call).

### Telemetry

A `:reviewer/scope-filter-applied` event fires whenever findings are moved, carrying `{:scope :filtered-count :kept-count :retry-path}`. Operators get visibility into when the structural backstop actually saved a verdict from prompt drift — useful for measuring whether the scope-aware prompt is doing its job or whether the filter is carrying the load.

## Test plan

- [x] `bb pre-commit` — 323 tests / 1185 assertions / 0 failures
- [x] `in-scope-issue?-test` — prefix / exact / boundary / trailing-slash / nil-scope / nil-file cases
- [x] `partition-issues-by-scope-test` — partition shape + nil-scope passthrough
- [x] **`test-reviewer-filters-out-of-scope-issues-before-verdict`** — full `core/invoke` integration: LLM emits both an in-scope blocker and an adjacent-file blocker; asserts that after the filter only the in-scope one drives `:review/issues` and the adjacent one is preserved in `:review/out-of-scope-observations`
- 53 reviewer tests / 206 assertions / 0 failures

### DIFF
--- a/components/agent/src/ai/miniforge/agent/reviewer.clj
+++ b/components/agent/src/ai/miniforge/agent/reviewer.clj
@@ -399,6 +399,43 @@
         (clean-scope-paths (:task/files-in-scope input))
         (when (map? intent) (clean-scope-paths (:scope intent))))))
 
+(defn- normalize-scope-entry
+  "Strip trailing slashes so a directory prefix matches `dir/file.clj` cleanly
+   without a false-positive on `dir-suffix/x.clj`."
+  [s]
+  (str/replace s #"/+$" ""))
+
+(defn in-scope-issue?
+  "True when a `ReviewIssue`'s `:file` falls inside `scope`.
+
+   - `scope == nil`  → true (no filtering; legacy behavior).
+   - `:file` blank   → true (file-level concern with no path attribution;
+                       conservatively kept so the LLM can still surface a
+                       real concern when it didn't name a file).
+   - Otherwise       → at least one scope entry equals the file or is a
+                       parent directory of it (entry + `/` is a prefix of
+                       the file path)."
+  [scope issue]
+  (let [file (some-> (:file issue) str str/trim not-empty)]
+    (cond
+      (nil? scope) true
+      (nil? file)  true
+      :else (boolean
+              (some (fn [entry]
+                      (let [trimmed (normalize-scope-entry entry)]
+                        (or (= file trimmed)
+                            (str/starts-with? file (str trimmed "/")))))
+                    scope)))))
+
+(defn partition-issues-by-scope
+  "Split `issues` against `scope` and return
+   `{:in-scope [...] :out-of-scope [...]}`. When `scope` is nil every
+   issue is in-scope (no filtering)."
+  [scope issues]
+  (let [{in true out false} (group-by #(in-scope-issue? scope %) issues)]
+    {:in-scope     (vec in)
+     :out-of-scope (vec out)}))
+
 (defn build-review-prompt
   "Construct the user prompt for LLM review from task data."
   [input]
@@ -964,11 +1001,30 @@
                                            timeout-only-review? nil
                                            parse-failed? :rejected
                                            llm-review (normalize-llm-decision (:review/decision llm-review)))
-                    initial-llm-issues    (get llm-review :review/issues [])
+                    ;; Resolve the task's scope once; partitioning happens
+                    ;; below at both the initial-parse and enumeration-retry
+                    ;; sites. nil means no filtering (legacy specs without
+                    ;; :scope continue to review every file).
+                    review-scope (effective-review-scope input)
+
+                    initial-llm-issues-raw (get llm-review :review/issues [])
+                    {initial-llm-issues          :in-scope
+                     initial-issues-filtered-out :out-of-scope}
+                    (partition-issues-by-scope review-scope initial-llm-issues-raw)
+
                     initial-llm-strengths (get llm-review :review/strengths [])
                     initial-llm-summary   (:review/summary llm-review)
-                    initial-llm-out-of-scope (sanitize-review-issues
-                                               (get llm-review :review/out-of-scope-observations))
+                    initial-llm-out-of-scope (into
+                                               (sanitize-review-issues
+                                                 (get llm-review :review/out-of-scope-observations))
+                                               initial-issues-filtered-out)
+                    _ (when (seq initial-issues-filtered-out)
+                        (log/info logger :reviewer
+                                  :reviewer/scope-filter-applied
+                                  {:data {:scope             review-scope
+                                          :filtered-count    (count initial-issues-filtered-out)
+                                          :kept-count        (count initial-llm-issues)
+                                          :retry-path        :initial}}))
 
                     ;; ENUMERATION VALIDATOR: a rejection without enumerated
                     ;; :blocking findings is malformed — the implementer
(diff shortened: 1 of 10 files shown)


components/agent/src/ai/miniforge/agent/reviewer.clj
```diff
@@ -964,11 +1001,30 @@
...
-                    initial-llm-out-of-scope (sanitize-review-issues
-                                               (get llm-review :review/out-of-scope-observations))
+                    initial-llm-out-of-scope (into
+                                               (sanitize-review-issues
+                                                 (get llm-review :review/out-of-scope-observations))
+                                               initial-issues-filtered-out)
+                    _ (when (seq initial-issues-filtered-out)
+                        (log/info logger :reviewer
+                                  :reviewer/scope-filter-applied
+                                  {:data {:scope             review-scope
```
H31MDALLR: We have this map, including the entire map under the data key, constructed more than once. That means that we should be extracting this to a factory function.

components/agent/src/ai/miniforge/agent/reviewer.clj
```diff
@@ -1009,8 +1078,10 @@
...
                     llm-out-of-scope (if recovered-review
-                                       (sanitize-review-issues
-                                         (get recovered-review :review/out-of-scope-observations))
+                                       (into
+                                         (sanitize-review-issues
+                                           (get recovered-review :review/out-of-scope-observations))
+                                         recovered-issues-filtered-out)
                                        initial-llm-out-of-scope)
 
                     ;; Merge decisions: gates can override LLM
```
H31MDALLR, line 1026: This file is well over a thousand lines. Decompose it. The function also looks like it's enormous. It should probably be decomposed. I can't see the entirety of it because we only changed a piece.

components/agent/test/ai/miniforge/agent/reviewer_test.clj
```diff
@@ -1184,3 +1184,120 @@
...
+            result   (core/invoke reviewer {} input)
+            review   (:artifact result)
+            issues   (vec (:review/issues review))
+            out      (vec (:review/out-of-scope-observations review))]
+        (is (= 1 (count issues))
+            "only the in-scope blocker remains as a verdict-driving issue")
+        (is (= "in-scope blocker" (:description (first issues))))
+        (is (= 1 (count out))
+            "the adjacent-file blocker is preserved as an advisory observation")
+        (is (= "adjacent-file blocker" (:description (first out))))))))
```
H31MDALLR, line 1197: This file is also 1,300 lines long. Decompose this into multiple test files. Make sure we have coverage for the ginormous reviewer namespace and, of course, the decomposed namespaces after you finish decomposition.

components/agent/src/ai/miniforge/agent/reviewer.clj
```diff
@@ -399,6 +399,43 @@
         (clean-scope-paths (:task/files-in-scope input))
         (when (map? intent) (clean-scope-paths (:scope intent))))))
 
+(defn- normalize-scope-entry
+  "Strip trailing slashes so a directory prefix matches `dir/file.clj` cleanly
+   without a false-positive on `dir-suffix/x.clj`."
+  [s]
+  (str/replace s #"/+$" ""))
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 93c433b3 — the helper moved to the new `reviewer.scope` ns and the docstring now correctly attributes the boundary check to `entry-matches-file?` (extracted from the inline anon-fn per rule 210). `normalize-scope-entry`'s job is the `dir//file` prevention you described.

components/agent/test/ai/miniforge/agent/reviewer_test.clj
```diff
@@ -1184,3 +1184,120 @@
...
+                                   :file "bases/cli/src/main.clj"
+                                   :description "x"})))
+
+  (testing "boundary check — similar prefix is NOT a match"
+    ;; Would be a false positive without the trailing-`/` normalization;
+    ;; "components/foo" is NOT a parent of "components/foobar/x.clj".
+    (is (not (reviewer/in-scope-issue? sample-scope
+                                       {:severity :blocking
+                                        :file "components/foobar/x.clj"
+                                        :description "x"}))))
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 93c433b3 (moved to `reviewer/scope_test.clj:in-scope-issue?-test`). The boundary test now uses scope entry `["components/foo"]` against file `"components/foobar/x.clj"` so the assertion actually exercises the trailing-`/` boundary; the prior `components/foo/src` vs `components/foobar/x.clj` would have passed even with the boundary removed.

components/agent/src/ai/miniforge/agent/reviewer.clj
H31MDALLR: Fixed in 93c433b3 — extracted to `reviewer.scope/filter-applied-event` which builds the canonical `{:data {:scope :filtered-count :kept-count :retry-path :message}}` payload (now also carrying a localized message). Called from both the initial-parse and enumeration-retry filter sites in reviewer.clj. Covered by a `filter-applied-event-test` in scope_test.clj.

components/agent/src/ai/miniforge/agent/reviewer.clj
H31MDALLR: Partially addressed in 93c433b3 — first decomposition step extracted the scope code into `reviewer.scope` (136 lines), bringing reviewer.clj from 1290→1229. Full decomposition (gates, prompts, llm-response, artifact builders, and the giant `create-reviewer` invoke-fn closure you flagged) is queued as 4 follow-up PRs since each is its own sizable extraction. Want me to roll those into THIS PR before merge, or land them as a focused stack? Happy to do either — wanted the migration + comment fixes in first so the scope contract is solid before the bigger reorg.

components/agent/test/ai/miniforge/agent/reviewer_test.clj
H31MDALLR: Partially addressed in 93c433b3 — extracted scope-specific pure-helper tests to `reviewer/scope_test.clj` (169 lines, includes a `review-issue` factory per the testing-standards repeated-data rule). Reviewer-side wiring tests (prompt rendering + end-to-end `core/invoke` integration) stay in `reviewer_test.clj` since they exercise reviewer.clj's wiring. Reviewer_test.clj is down to 1197 lines. Companion test-file decompositions will mirror the source-side follow-up PRs referenced on comment #3360249198.

PR #1189, opened 2026-06-16
Wire loop lifecycle event emissions in monitor_loop.clj.

H31MDALLR:
---
miniforge-workflow: ece151ed-4b99-45bd-b56b-f28321a0d499
spec: work/monitor-loop-event-emission.spec.edn
task: 11111111-1111-1111-1111-111111111111
commit: 66c5f410da14f9e331366976d8f418c63c530d77
generated-by: miniforge
---

## Summary

Wire loop lifecycle event emissions in monitor_loop.clj.

## Changes

- `components/pr-lifecycle/src/ai/miniforge/pr_lifecycle/monitor_loop.clj` (modify)

## Review

**Decision**: approved

All three emission sites are wired correctly. loop-started fires after log-loop-start! and before step-monitor-loop!. loop-stopped fires after step-monitor-loop! returns, with (get @monitor :stop-reason :completed) covering no-open-prs, manual-stop, and the natural-completion default in a single emission point. cycle-completed with nil pr-number fires after the swap! in finalize-loop-iteration!, correctly distinguished from per-PR cycle-completed events by the nil field. The stop-reason writes are complete: :no-open-prs in stop-when-no-open-prs!, :manual-stop in stop-monitor-loop. Verification findings are accurate: event-type-registry exists as an SSE/browser layer (inapplicable here, existing constructors already registered in monitor-events), and the event-bus is confirmed as a side bus (follow-up required but correctly deferred). One in-scope warning: swap! return value should be captured rather than re-dereferencing @monitor. No blocking in-scope issues.

### Known issues (non-blocking)

Merged with 2 unresolved non-blocking issue(s) recorded for follow-up:

- `components/pr-lifecycle/src/ai/miniforge/pr_lifecycle/monitor_loop.clj:163` — TOCTOU: `swap!` return value is discarded; `:cycles` is then read from a second `@monitor` dereference. `(swap! monitor ...)` returns the new state map atomically — reading `(:cycles @monitor)` after it is a separate, non-atomic read. Another writer could swap the atom between the two calls. In this component's single-threaded loop this is practically safe, but the pattern is wrong and will silently emit a stale iteration count if concurrency is ever introduced.
- `components/pr-lifecycle/src/ai/miniforge/pr_lifecycle/monitor_loop.clj:20` — Rule 009 (self-documenting-code): The VERIFICATION FINDINGS comment block narrates investigation reasoning — bus identity, registry lookup, conclusions. Per rule 009, source files describe the current contract; investigation narrative belongs in the PR doc or a follow-up issue, not the source. 'Never narrate your own reasoning in the code. The diff and PR description are for that.' The spec said document findings; the right location is the PR doc, not a code comment block.

## Test plan

- [ ] Verify changes compile and tests pass
- [ ] Review generated code for correctness

---
Generated by Miniforge SDLC workflow (1 file)

### DIFF
--- a/components/pr-lifecycle/src/ai/miniforge/pr_lifecycle/monitor_loop.clj
+++ b/components/pr-lifecycle/src/ai/miniforge/pr_lifecycle/monitor_loop.clj
@@ -29,6 +29,27 @@
    [ai.miniforge.pr-lifecycle.pr-poller :as poller]
    [ai.miniforge.schema.interface :as schema]))
 
+;; VERIFICATION FINDINGS (2026-06-15)
+;;
+;; (A) event-type-registry.edn
+;;     Exists at components/event-stream/resources/config/event-stream/event-type-registry.edn.
+;;     It is the SSE/browser-layer registry for the event-stream component.
+;;     The :pr-monitor/* types used here (loop-started, loop-stopped, cycle-completed) are
+;;     NOT in that registry — nor need they be. They are registered in
+;;     monitor_events/monitor-event-types (the in-component source of truth for this
+;;     namespace group). No new event types are introduced in this PR; all three
+;;     constructors already existed in monitor_events.clj.
+;;
+;; (B) event-bus identity
+;;     The :event-bus value threaded through monitor.config is created by
+;;     events/create-event-bus, which is an in-memory atom
+;;     {:events [] :subscribers {} :filters {}}.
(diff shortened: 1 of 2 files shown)


H31MDALLR:
Superseded by #1211, which consolidates the loop-event-emission feature + tests onto current main (this stack was based on stale pre-#1197/#1198/#1206/#1209 main). The 5 code review comments from #1189 are addressed in #1211.

PR #1236, opened 2026-06-20
feat: typed handoff acts (phase outcomes, meta-loop halt, PR-monitor decision)

H31MDALLR:
## Overview

Makes three SDLC handoff points explicit, typed acts instead of signals reconstructed from status flags or FSM transitions — the one idea borrowed from the SGF/MiniForge ontology review (typed acts), implemented by extending miniforge's own closed enums rather than importing a generic performative vocabulary.

1. **Phase outputs** — `:phase/outcome` gains `:blocked` (refusal + machine-readable cause) and `:redirected` (request to the pipeline).
2. **Meta-loop halt** — new `:meta-loop/halt-requested` event: a meta-agent halt is now first-class on the stream, not just a value in the coordinator result.
3. **PR-monitor decision** — new `:pr-monitor/decision-recorded` bus event recording the classify→act choice (`:fix`/`:answer`/`:approve`/`:skip`) with provenance, at the single routing point.

A shared closed `RefusalReason` enum backs both phase `:blocked` and meta-loop halts. Also closes a pre-existing gap: the `:agent/message-sent` / `:agent/message-received` constructors had no schema; both are now defined.

## Design notes

- **`:status` not collapsed.** It stays the 2-valued FSM control flag (read by many leave-handlers); `:phase/outcome` carries the act vocabulary on the observed layer only.
- **PR-monitor decision is a bus event, not a stream event.** The monitor runs on its own bus, decoupled from the observability stream by design. Surfacing it into evidence bundles needs a pr-lifecycle→event-stream bridge — flagged as follow-on.

## Layering

Four commits, one per stratum (contract → phase → meta-loop → pr-monitor), each under the 200-line commit budget. Kept as one PR because the pieces are not independently mergeable in isolation (an emitter referencing an undefined enum value fails; the enum with no emitter is dead). Meaningful change 340 lines, under the 400 PR limit.

## Verification

clj-kondo clean; affected tests green; `clojure -M:poly check` OK; full pre-commit hook (poly check, lint, fmt, smoke, GraalVM) passed on every commit.

Full detail: `docs/pull-requests/2026-06-19-chris-typed-phase-acts.md`.

### DIFF
--- a/bases/cli/src/ai/miniforge/cli/workflow_runner/display.clj
+++ b/bases/cli/src/ai/miniforge/cli/workflow_runner/display.clj
@@ -114,12 +114,16 @@
                                                             {:phase phase}))
       :workflow/phase-completed (let [outcome (or (:phase/outcome event) :completed)
                                       color   (case outcome
-                                                :failure :red
-                                                :skipped :yellow
+                                                :failure    :red
+                                                :blocked    :red
+                                                :skipped    :yellow
+                                                :redirected :yellow
                                                 :green)
                                       symbol  (case outcome
-                                                :failure "✗"
-                                                :skipped "○"
+                                                :failure    "✗"
+                                                :blocked    "⊘"
+                                                :skipped    "○"
+                                                :redirected "↻"
                                                 "✓")]
                                   (str (colorize color (messages/t :workflow-runner/phase-completed
                                                                    {:symbol symbol
--- a/components/event-stream/src/ai/miniforge/event_stream/core.clj
+++ b/components/event-stream/src/ai/miniforge/event_stream/core.clj
@@ -523,6 +523,7 @@
         (cond->
           (:duration-ms result) (assoc :phase/duration-ms (:duration-ms result))
           (:review-decision result) (assoc :phase/review-decision (:review-decision result))
+          (:phase/blocked-reason result) (assoc :phase/blocked-reason (:phase/blocked-reason result))
           (:artifacts result) (assoc :phase/artifacts (:artifacts result))
           (:error result) (assoc :phase/error (:error result))
           request (assoc :phase/transition-request request)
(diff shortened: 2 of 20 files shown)


H31MDALLR:
## Tracked debt: collapse `:status` / `:phase/outcome` into one control vocabulary

This PR types the *observed* layer (`:phase/outcome` on the event stream + evidence) but deliberately leaves the *internal* control flag, `:status`, alone. Recording the reason and the end-state so the deferral is explicit, not silent.

**Current state.** A phase's `:phase/outcome` is derived from the phase-result `:status` (`:success` / `:error` / `:failed`) plus side markers. `:blocked` has no status of its own — it rides on `:status :error` + a `:phase/blocked-reason` key. So internally the act is still inferred from field presence, the thing this PR removes on the observed layer.

**Why it's debt, not style.** The relation between the act and the control status is now written in three places:

1. `phase-result/blocked` (`components/phase/.../phase_result.clj`) — act → `:error` + marker key
2. `runner-events/phase-outcome` (`components/workflow/.../runner_events.clj`) — `:error` + marker → `:blocked`
3. `tui-views/phase-status` (`components/tui-views/.../persistence.clj`) — `:blocked` → `:failed`

(plus presentation re-mapping in `bases/cli/.../workflow_runner/display.clj`). Same fact, multiple translations, multiple drift points.

**End-state.** One closed control vocabulary that *is* the act: the workflow FSM, the phase leave-handlers, and the emitted event all read the same enum. `:phase/outcome` becomes a pass-through, or replaces `:status`; the `phase-outcome` derivation and the consumer re-maps collapse.

**Why deferred here.** Blast radius: every phase leave-handler and the workflow FSM dispatch on `:status` across all phases. Focused refactor with its own test surface — not a rider on a vocabulary change. Tracked as follow-up.

PR #1239, opened 2026-06-20
Policy enforcement program: gap analysis, judge eval, locked config (docs + eval infra)

H31MDALLR:
Program to honor the trust boundary — *every defined policy is applied; nothing that violates a policy makes it through the workflow.* This PR is the **analysis + measurement + decision** layer (docs + eval harness). No production behavior changes yet; the implementation (E1) is scoped here for follow-up PRs.

Full detail: `docs/research/policy-enforcement-gap-analysis.md`.

## Findings — current enforcement is "soft by construction"
- 58 rules; **only 1 deterministically blocks** (`clojure-no-requiring-resolve`).
- The semantic policy gate is **dormant in production**: `:complete-fn` is set nowhere outside tests, so `with-semantic-wiring` never injects the judge and all 54 `:custom`/`:semantic` rules silently pass. Fail-open.
- Container-**security** rules (`runtime-*`) are non-blocking warnings.
- `enforcement.action` is a mechanical severity default (`mdc_compiler.clj`), never a per-rule policy-vs-guidance decision.

## Decisions recorded (from the program owner)
- Concrete rule ⇒ policy ⇒ hard-halt; detector reliability is handled by evidence + ratchet, not by demoting the rule.
- Security → hard-halt immediately; add a defense-in-depth + AI-security pack.
- Semantic scope = changed files only; in-scope files must comply (fix-and-gate).
- Caching is backend-conditional (don't convert free subscription quota into metered spend).

## Judge eval (`eval/policy-fidelity/`)
A seeded-synthetic harness that scores judge strategies × models against known truth. Two contamination bugs were found and fixed by hardening (answer-key leakage in fixtures; a hand-oracle less thorough than the judge; intent-revealing filenames).

**Locked judge config: `claude-opus-4-8` + batched (one call per changed file), fail-closed, changed-files-scoped.** On the grown corpus (12 fixtures, 9 rules, 3 trials): **recall 0.95, precision 0.92, zero parse-fails.** Batched is parse-fail-unreliable below Opus (model-bound). An LLM judge is ~95% recall — strong, not infallible — so it's one layer in defense-in-depth, not an oracle.

**Cost** (`cost_model.clj`, analytic): opus-B ≈ $1.03/gate uncached → **~$0.22 with cross-run rule-pack caching** (the compiled pack is a write-once-read-forever prefix; per-gate cost becomes the changed files, not the policy set). Caching auto-enables on `cache_control`-capable API/BYOM backends; no-op (plan-covered) on the subscription CLI.

## Scoped for follow-up (E1+)
Wire the judge (synthesize/enable semantic detection; API/HTTP backend with `cache_control` gated on backend capability; fail-closed; changed-files scope; guidance tier) → per-rule action classification in the `standards/miniforge` submodule → security pack → ratcheted fix-violation waves.

Note: eval corpus is synthetic and modest — grow toward real agent-output samples before flipping subtle semantic rules to hard-halt.

### DIFF
--- a/eval/policy-fidelity/fixtures/seed/order_service.clj.txt
+++ b/eval/policy-fidelity/fixtures/seed/order_service.clj.txt
@@ -1,0 +1,10 @@
+(ns eval.seed.order-service
+  (:require [ai.miniforge.payments.interface :as payments]))
+
+(defn charge-order
+  "Charge an order. Returns the payment result."
+  [order]
+  (let [result (payments/charge (:amount order))]
+    (if (= :error (:status result))
+      (throw (ex-info "charge failed" {:order order}))
+      (assoc order :charged? true :tier "gold" :loyalty-days 90))))
--- a/eval/policy-fidelity/run.clj
+++ b/eval/policy-fidelity/run.clj
@@ -37,10 +37,23 @@
 (def fixture-root "eval/policy-fidelity/fixtures")
 (def truth (edn/read-string (slurp (io/file fixture-root "truth.edn"))))
 
+(defn ->rule-kw
+  "Coerce a model-emitted :rule-id to a keyword — accepts a keyword, a
+   \":std/foo\" string, or a \"std/foo\" string. nil for anything else."
+  [x]
+  (cond (keyword? x) x
+        (string? x)  (keyword (str/replace x #"^:" ""))
+        :else        nil))
+
 (defn load-candidate-rules []
   (let [by-id (into {} (map (juxt :rule/id identity))
-                    (into (ab/load-builtin-rules) (ab/load-standards-rules)))]
-    (mapv by-id (:candidate-rules truth))))
+                    (into (ab/load-builtin-rules) (ab/load-standards-rules)))
+        ids (:candidate-rules truth)
+        missing (remove by-id ids)]
+    (when (seq missing)
+      (throw (ex-info (str "candidate rules missing from packs: " (vec missing))
+                      {:missing (vec missing)})))
+    (mapv by-id ids)))
 
 (defn bounded-pmap [n f coll]
   (vec (mapcat (fn [batch] (mapv deref (mapv #(future (f %)) batch)))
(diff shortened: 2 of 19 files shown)


eval/policy-fidelity/fixtures/seed/order_service.clj.txt
```diff
@@ -0,0 +1,10 @@
+(ns eval.seed.order-service
+  (:require [ai.miniforge.payments.interface :as payments]))
+
+(defn charge-order
```
H31MDALLR, line 4: What is this? Why do we have payment processing in miniforge? It's also incorrect. It uses raw strings and the tiers are defined as strings as well, and LoyaltyDays is defined as an immediate value, so this entire function is a nice set of standards violations. In addition to being, why the hell do we have a payment service in miniforge?

H31MDALLR: Test collateral, not a component. `order_service.clj.txt` is a seeded-synthetic fixture for the policy-judge fidelity eval (`eval/policy-fidelity/`), not real code: stored as `.clj.txt` (not `.clj`) so it is not a namespace, not on any source path, skipped by clj-kondo, never compiled or loaded. The standards violations are deliberate — it is the judge's exam question; `truth.edn` is the answer key listing exactly which rules each fixture violates. The payment/order domain is arbitrary filler to give the judge non-trivial code.

Marked it as such in `587d0922`: added `eval/policy-fidelity/fixtures/seed/README.md` stating plainly that the dir is test collateral, violations are intentional, and in-file violation markers are forbidden (they leak the answer key to the judge — that leak is why I removed them earlier). Filenames are kept domain-neutral on purpose: a `*_fixture`/`*_test`/`near_miss`/`clean_*` name primes the judge via the path it sees and corrupts the precision measurement. So the marker is the README + the `fixtures/seed/` location + the `.txt` extension, not the filename.

eval/policy-fidelity/run.clj
```diff
@@ -0,0 +1,185 @@
...
+       :violations (vec (filter map? v))})))
+
+;; ---- scoring ----
+(defn confusion [rows]               ; rows: {:seeded? :flagged?(true/false/:parse-fail)}
+  (let [hit? (fn [r] (true? (:flagged? r)))
+        pf?  (fn [r] (= :parse-fail (:flagged? r)))]
+    {:tp (count (filter #(and (:seeded? %) (hit? %)) rows))
+     :fn (count (filter #(and (:seeded? %) (not (hit? %))) rows))
+     :fp (count (filter #(and (not (:seeded? %)) (hit? %)) rows))
+     :tn (count (filter #(and (not (:seeded? %)) (not (hit? %)) (not (pf? %))) rows))
```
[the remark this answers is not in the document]
H31MDALLR: False positive — the parens are balanced and the script loads and runs (three full eval passes produced `results/latest.edn`). The construct flagged is the anonymous-function reader macro `#( ... )`; `#(and (not (:seeded? %)) (not (hit? %)) (not (pf? %)))` is one complete reader form, then ` rows)` closes `filter` and `)` closes `count`. A paren-counter that does not treat `#(` as an opening reader macro will miscount here. No imbalance. (The function was revised in this push anyway, to exclude backend-error rows from the matrix per the run.clj:101 comment.)

PR #1490, opened 2026-07-25
fix(workflow): preserve merge-anomaly message through dag/err

H31MDALLR:
## Summary
- `run-mini-workflow` (dag_task_execution.clj) short-circuits a v2 multi-parent merge failure by putting the full typed anomaly map on `:error`, not a string.
- `workflow-result->dag-result` passed that map straight into `dag/err`'s message slot, so `unwrap-anomaly` (which requires a string `:message`) silently fell back to its generic "Unwrap called on error result" text instead of the anomaly's own message — losing the merge-failure detail for any caller.
- Fix: extract `:anomaly/message` for the message slot when `:error` is a merge anomaly (`dag-merge-anomaly/merge-error?`), and keep the full anomaly map reachable under `:merge/anomaly` in the error's `:data` so downstream consumers (dashboard, evidence bundle) can still classify by `:anomaly/category` per the existing `run-mini-workflow` docstring contract.
- Pre-existing bug, unchanged since before this branch's `dag_orchestrator.clj` → 12-namespace split (identical logic in the monolithic file); this PR fixes it in the split shape before #1485 merges.

Stacked on #1485 (`fix/dag-orchestrator-namespace-split`) since the target file only exists post-split.

## Test plan
- [x] New unit tests in `dag_task_execution_test.clj`: string-error path unchanged, merge-anomaly path unpacks `:anomaly/message` and preserves the full map under `:merge/anomaly`.
- [x] Read `merge_parent_branches_integration_test.clj` and `dag_resilience_execution_test.clj` — neither exercises `workflow-result->dag-result` directly, no conflicting expectations.
- [x] Ran the 8 DAG-related workflow test namespaces (incl. both files above): 96 tests / 310 assertions, 0 failures/errors.
- [x] `bb lint:stratum` + pre-commit hooks (lint, poly check, smoke tests, GraalVM compat) all passed on commit.


H31MDALLR:
@copilot resolve the merge conflicts on this branch.

PR #327, opened 2026-03-21
Fix artifact persistence, tool tracking, and control plane wiring

H31MDALLR:
## Summary

- **Nil-output guard**: implement phase now treats `{:success true, :output nil}` as error and retries, preventing release phase crashes
- **Tool-use tracking**: tool names extracted from Claude CLI stream-json events, propagated through LLM response chain, and logged for diagnostics
- **Proper logging**: replaced all stderr hacks (`binding [*out* *err*]`) with real logger throughout implement/release/implementer/llm-client
- **Multi-vendor control plane**: new adapter protocol, orchestrator, registry, decision queue, heartbeat watchdog, and miniforge/claude-code adapters
- **Classpath & test fixes**: wired 4 new components into deps.edn, removed aspirational tests referencing unimplemented functions, cleaned up accidentally committed worktree

## Test plan

- [x] All artifact persistence tests pass (5/5)
- [x] All release phase tests pass (9/9)
- [x] All implement phase tests pass (6/6)
- [x] All implementer agent tests pass (15/15)
- [x] All adapter-miniforge tests pass
- [x] All control-plane tests pass (compilation + registry/decision-queue/heartbeat)
- [ ] Remaining pre-existing CLI test failures (help text, fleet commands) are unrelated to this branch

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1642, opened 2026-08-05
feat: add incident provenance projection

H31MDALLR:
## Summary

- add an experimental read-only `build-dossier` API for Incident → Symbol/Range → Commit → PR projections
- attach explicit specification mappings, applicable policy rules, generated claims, evidence bases, coverage, and unresolved-source gaps
- pin repository revisions and preserve distinct range evidence and specification clauses without asserting `introduced-by` causality

## Validation

- `clojure -M:poly test brick:governance-provenance` — 8 tests, 40 assertions
- `bb pre-commit` — 338 smoke tests / 1,281 assertions and 8 GraalVM tests / 584 assertions
- real-repository dogfood resolved a pinned range through commit `cdb1034` to locally visible merge PR #327 and reported absent symbol/spec/policy sources as gaps

## Review notes

- The single-commit 200-line budget was explicitly overridden because source adapter, projection model, tests, and spec traceability form one usable read-only slice.
- Deliberately excludes remote provider queries, graph persistence, remediation authority, inferred symbols, and `introduced-by` claims.

See `docs/pull-requests/2026-08-05-feat-incident-provenance-projection.md` for the full handoff.

## PR budget

MINIFORGE_PR_BUDGET_OVERRIDE: 636 reportable lines, 36 over the 600 ceiling. The review-round-1 fix commit (normalize path separators before glob matching in mapping-matches?/applicable-rules; give blame-evidence an explicit :evidence/range-resolved? signal instead of a silently-nil range and a degenerate id) added 77 lines to a PR that was under budget before it, including two new tests pinning both fixes. Splitting a review-response fix out of the PR it fixes isn't a smaller slice, it's an artificial one -- this is the original read-only projection slice plus its own review fixes.


components/governance-provenance/src/ai/miniforge/governance_provenance/core.clj
```diff
@@ -0,0 +1,185 @@
...
+
+;------------------------------------------------------------------------------ Layer 0
+
+(defn ^{:stratum 0} empty-projection
+  []
+  {:nodes [] :edges [] :claims [] :paths [] :gaps []})
+
+(defn ^{:stratum 0} mapping-matches?
+  [path mapping]
+  (some #(policy/glob-matches? % path) (:mapping/file-globs mapping)))
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 0dbc27540 -- both mapping-matches? and applicable-rules now normalize \\ to / before matching (matches governance-provenance.git/valid-location?'s existing treatment of \\ as a separator). Added a direct test for it.

components/governance-provenance/src/ai/miniforge/governance_provenance/model.clj
```diff
@@ -0,0 +1,128 @@
...
+
+(defn ^{:stratum 2} blame-evidence
+  [revision subject facts commit]
+  (let [sha (:commit/sha commit)
+        range-n (range-node revision facts)
+        id (str "evidence:git-blame:" (:node/id subject) ":" (:node/id range-n) ":" sha)]
+    (node id :evidence revision {:source/type :git-blame :source/ref sha}
+          {:evidence/id id :evidence/type :git-blame
+           :evidence/subject (:node/id subject) :evidence/commit sha
+           :evidence/range (:node/content range-n)})))
```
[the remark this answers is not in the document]
H31MDALLR: Fixed in 0dbc27540 -- blame-evidence now uses a stable "unresolved-range" id segment instead of baking in nil (avoids the degenerate/collision-prone id), and adds an explicit :evidence/range-resolved? boolean so consumers don't have to infer "no range" from a missing map key. :evidence/range is only present when actually resolved. Added an end-to-end test (ls-tree failing, blame still resolving a commit) pinning this shape.

components/governance-provenance/src/ai/miniforge/governance_provenance/core.clj
```diff
@@ -0,0 +1,192 @@
...
+  (some? (:rule/id rule)))
+
+(defn ^{:stratum 0} commit-projection
+  [incident-id revision subject facts commit pr]
+  (let [commit-n (model/commit-node commit)
+        claim (model/attribution-claim incident-id subject commit)
+        evidence (model/blame-evidence revision subject facts commit)
+        pr-n (some-> pr model/pull-request-node)
+        nodes (cond-> [commit-n claim evidence] pr-n (conj pr-n))
+        edges (cond-> [(model/edge (:node/id subject) :changed-by (:node/id commit-n) :derived
```
H31MDALLR, line 50: We should have a common edge fn that should not need us to specify the keys (over and over again)

components/governance-provenance/src/ai/miniforge/governance_provenance/core.clj
```diff
@@ -0,0 +1,192 @@
...
+                  :when (and (nil? (:rule/revision rule)) (nil? (:rule/version rule)))]
+              {:gap/type :policy-rule-revision-unavailable :gap/rule (:rule/id rule)}))
+      :always
+      (into (for [commit commits
+                  :when (not (contains? pr-commits (:commit/sha commit)))]
+              {:gap/type :pull-request-unresolved :gap/commit (:commit/sha commit)})))))
+
+;------------------------------------------------------------------------------ Layer 2
+
+(defn ^{:stratum 2} project-location
```
H31MDALLR, line 166: This function is hard to read. The forms in the let block are long enough that they impair readability of this function. And you still essentially have methods in the composition to return the data, so this feels very messy. Even the refactor below doesn't feel coherent, which makes me wonder if this entire component needs to be rethought .. but it is more readable.

Example refactor:

```clojure


(defn range-edge [range-n subject]
  (let [range-node-id (:node/id range-n)
        subject-node-id (:node/id subject)
        same-id? (= range-node-id subject-node-id)]

    (if (and range-n (not same-id?))
      (model/edge subject-node-id
                  :cites range-node-id
                  :asserted [{:evidence/type :operator-selected-location}])
      {})))

(defn commits [incident-id revision subject facts]
  (mapv #(commit-projection incident-id
                            revision
                            subject
                            facts
                            %
                            (first (get prs-by-commit (:commit/sha %))))
        (:commits facts)))
(defn edges [incident-edge range-edge edges commits])
(defn nodes [subject range-n nodes commits])
(defn paths [commits incident-id]
  (mapv #(update % :path/nodes
                 (fn [nodes]
                   (into [(str "incident:" incident-id)] nodes)))
        (mapcat :paths commits)))

(defn project-location
  [incident-id revision facts mappings rules policy-context]
  (if-let [subject (model/subject-node revision facts)]
    (let [range-n (model/range-node revision facts)
          incident-edge (model/edge (str "incident:" incident-id)
                                    :implicates (:node/id subject)
                                    :asserted [{:evidence/type :operator-selected-location}])
          range-edge (range-edge range-n subject)
          governance (governance-projection subject
                                            (get-in facts [:location :path])
                                            mappings
                                            rules
                                            policy-context)
          prs-by-commit (group-by :pull-request/contains-commit (:pull-requests facts))
          commits (commits incident-id revision subject facts)
          paths (paths commits incident-id)
          {:keys [nodes edges]} governance]
      {:nodes (edges incident-edge range-edge edges commits)
       :edges (nodes subject range-n nodes commits)
       :claims (vec (mapcat :claims commits))
       :paths paths
       :gaps (location-gaps facts mappings rules policy-context)})
    (assoc (empty-projection) :gaps (location-gaps facts mappings rules policy-context))))
```

components/governance-provenance/src/ai/miniforge/governance_provenance/core.clj
H31MDALLR: Added `model/edge1` in #1660 -- every one of the 7 model/edge call sites in this component was hand-building the same single-entry `[{:evidence/type ... :evidence/ref ...}]` basis vector. edge1 takes the evidence map directly:

```clojure
(defn edge1 [from type to status evidence] (edge from type to status [evidence]))
```

Lives in model.clj next to edge, not core.clj -- keeping it in core.clj would've cascaded an extra stratum-lint layer through every caller. All 7 sites now go through it: components/governance-provenance/src/ai/miniforge/governance_provenance/core.clj@539e86d6b.

components/governance-provenance/src/ai/miniforge/governance_provenance/core.clj
H31MDALLR: Took a real pass at this in #1660 (2 commits) rather than a spot-fix.

First commit split project-location's caller, build-dossier -- same root problem, one function doing validate+collect+project+assemble with the whole result map built from inline computed expressions.

Second commit went at project-location itself: extracted location-implicates-edge and location-cites-edge as named single-purpose fns (previously two inline model/edge calls buried in the let), location-commit-projections (the mapv + its prs-by-commit grouping), and the edge1 helper from the other thread. commit-projection's three edge constructions are now named (changed-by-edge/supports-edge/contained-in-edge) instead of packed into one dense cond-> vector literal, and governance-projection's two edge-builders are named instead of anonymous.

I didn't go as far as your sketch (a top-level defn per let-binding) -- landed on named-fn-per-binding but still assembled in one let, since your own note that even that version 'doesn't feel coherent' read as genuine uncertainty about the shape, not a spec to match exactly. If it's still not there, or you want the fuller decomposition, say so on #1660 -- easier to iterate there than reopen a merged PR. Didn't resolve this thread myself; leaving that to you once you've looked at the diff.

PR #1672, opened 2026-08-06
fix: return anomalies from durable effect boundaries

H31MDALLR:
## Layer
Infrastructure

## Depends on
- None

## What this adds
Returns serialization and validation failures from durable effect boundaries as anomaly values before any write.

## Strata affected
- effect-transaction persistence
- execution-grant breach history

## Validation
- Normal pre-commit checks
- Focused affected-component tests

### DIFF
--- a/components/effect-transaction/test/ai/miniforge/effect_transaction/store_test.clj
+++ b/components/effect-transaction/test/ai/miniforge/effect_transaction/store_test.clj
@@ -186,3 +186,18 @@
     (is (= original (store/create! dir original)))
     (is (anomaly/anomaly? (store/create! dir replacement)))
     (is (= original (fx/read-record dir (:effect/id original))))))
+
+(deftest ^{:stratum 2} lifecycle-writes-propagate-store-anomalies-test
+  (let [dir (tmp-dir)
+        write-failure (anomaly/anomaly :unavailable "disk unavailable" {})]
+    (with-redefs [store/create! (fn [& _] write-failure)
+                  store/transition! (fn [& _] write-failure)]
+      (is (= write-failure
+             (effect-record/propose!
+              dir
+              {:effect-class :effect/merge
+               :grant-id (random-uuid)
+               :envelope-id (random-uuid)}
+              now)))
+      (is (= write-failure
+             (effect-record/advance! dir (record) {:effect/state :committing} now))))))
--- a/docs/pull-requests/2026-08-05-codex-standards-effect-grant-anomalies.md
+++ b/docs/pull-requests/2026-08-05-codex-standards-effect-grant-anomalies.md
@@ -1,0 +1,35 @@
+<!--
+  Title: Miniforge.ai
+  Author: Christopher Lester (christopher@miniforge.ai)
+  Copyright 2025-2026 Christopher Lester. Licensed under Apache 2.0.
+-->
+# fix: return anomalies from durable effect boundaries
+
+## Overview
+
+Moves effect-transaction and execution-grant persistence validation failures onto the anomaly value path.
+
+## Changes in Detail
+
+- Reject unsupported timestamps before writing.
+- Propagate persistence anomalies to callers.
+- Update focused tests for the returned anomaly contract.
+
+## Testing Plan
+
+- Focused effect-transaction and execution-grant tests
+- Normal pre-commit validation
+
+## Deployment Plan
+
+No migration or rollout is needed.
+
+## Related Issues/PRs
+
+- Base Branch: `main`
+- Depends On: none
+
+## Checklist
+
+- [x] Audit gap fixed
+- [x] Pre-commit checks passed


H31MDALLR:
Closing: this PR no longer changes any behavior.

**What is on the branch now** (`949ca1a`, +50/−0): one test (`lifecycle-writes-propagate-store-anomalies-test`) and the PR doc. No production code.

**What it used to contain.** The review history here shows 8 changed files. The pre-rebase branch was:

```
8a098bf fix: propagate rebased effect write failures        (record.clj + test)
7361cca fix: clarify durable anomaly contracts              (3 production files)
f03764c docs: document durable effect anomaly remediation
9e3c2fe fix: return anomalies from durable effect boundaries (6 files, +119/-84)
```

The current branch keeps only the doc commit and a gutted `949ca1a`: `9e3c2fe` and `7361cca` were dropped entirely, and `949ca1a` lost the `record.clj` hunk that `8a098bf` carried. The rebase conflicted because the fix patched `store/write!`, which #1705 (`e67c640`) removed when it split `store.clj` into `persistence.clj` + `codec.clj`. The conflicts were resolved by discarding the changes rather than porting them.

**The intent is already satisfied on `main`** — which is why CI is green on a diff containing no production code:

- `record/propose!` and `record/advance!` validate against the schema before writing and return `store/create!` / `store/transition!` in tail position, so persistence anomalies propagate to callers.
- `persistence/persist!` wraps serialization in `try/catch → write-failure`, and `pr-str` is evaluated before `spit`, so a rejected record never reaches disk.
- `interface/record-breach!` validates `schema/Breach` before writing, and `interface/revoke-for-cause!` validates `[ExecutionGrant BreachObservation inst?]`, which makes the constructed breach valid by construction.
- `breach_test.clj` already carries `malformed-breach-is-refused-test`, `storage-failures-are-returned-as-data-test`, and `both-inst-types-round-trip-test`.

The surviving test asserts behavior that already shipped, so it passes without any production change.

The reason to close rather than merge is the doc: it claims "Reject unsupported timestamps before writing", lists execution-grant breach history under **Strata affected** (nothing under `components/execution-grant/` is touched), and ticks "Audit gap fixed". Merging would put an inaccurate entry into `docs/pull-requests/`, which is the provenance trail for this repo.

No follow-up PR is needed — every write path into both durable stores is guarded at the interface boundary. If the regression test is worth keeping on its own merit, it can be reopened as a test-only PR with a doc that says so.

---
_Generated by [Claude Code](https://claude.ai/code)_

H31MDALLR:
Correction to my previous comment, for the record.

This PR's content **was already on `main`** when I closed it. Commit `3d2361e4` ("fix: return anomalies from durable effect boundaries (#1672)", 2026-08-11) squash-merged it, and its two files are byte-identical to head `949ca1a`. The regression test lives at `store_test.clj:190` today. GitHub left the PR in `open` state because the squash landed as a direct push to `main` rather than through the merge button.

So closing was correct, but the reason is "already merged and stale", not "never landed". My suggestion that the test could be reopened as a test-only PR is moot — it is already there.

The rest of the analysis stands, and the timeline is:

- **Aug 5–6** — PR carried 8 files, including the `store.clj` / `core.clj` / `breach.clj` / `eligibility.clj` fix.
- **Aug 7** — rebase onto post-#1705 `main` dropped every production file (they patched `store/write!`, which #1705 had removed), leaving the test and doc.
- **Aug 11** — the reduced two-file version was squash-merged as `3d2361e4`.
- **Aug 22** — closed here.

One consequence worth flagging: because the reduced version merged, `docs/pull-requests/2026-08-05-codex-standards-effect-grant-anomalies.md` is now on `main` still claiming "Reject unsupported timestamps before writing", listing execution-grant breach history under **Strata affected**, and ticking "Audit gap fixed" — none of which the merged diff does. The provenance trail overstates that commit. Correcting that doc in place would be a reasonable follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

H31MDALLR:
One factual correction to both comments above, since this thread is part of the provenance trail.

Both credited the durable-store restructure to #1705 (`e67c640`). That attribution is unsupported. `e67c640` is a **root commit** with no parents — the repository's visible history begins there:

```
$ git log -1 --format="%P" e67c640      # (empty — no parents)
```

It carries the entire repository tree, and it already contains `persistence.clj` and `codec.clj` as part of that initial tree, not as a change it made. `git log -- <path>` named it for these files only because history *stops* there; that is truncation, not authorship. The split predates everything now visible, so no commit in this history can be credited with it.

What this does not change: `store/write!` does not exist on `main`, the 2026-08-07 rebase dropped the six production files because they patched it, and `main` supplies the anomaly guarantees by other means — the tail-position returns in `record.clj`, the `try`/`catch` in `persistence/persist!`, and the schema validation in `interface/record-breach!` and `interface/revoke-for-cause!`. Those were each verified directly against the code and stand as written.

The corrected wording is in #1828 (`c6cb96e`).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1685, opened 2026-08-07
fix(bench): isolate the bench sandbox from the checkout it was launched from

H31MDALLR:
## Overview

Provision the bench/dogfood sandbox as an independent clone, and add a fail-closed guard that refuses a bench sharing a git common dir with the launching checkout.

Full write-up: [docs/pull-requests/2026-08-06-bench-sandbox-isolation.md](docs/pull-requests/2026-08-06-bench-sandbox-isolation.md)

## The defect

On 2026-08-06 a bench run left the live miniforge checkout with `remote.origin.url` pointing at `~/.miniforge/bench/origin.git` and `refs/remotes/origin/main` force-updated **backwards** to that mirror's stale head. Branches cut from `origin/main` afterwards were based on a stale commit and pushed to the local mirror, so the PRs never appeared on GitHub.

Root cause: `~/.miniforge/bench/repo` was a linked `git worktree` of the live checkout.

```
$ git -C ~/.miniforge/bench/repo rev-parse --git-dir --git-common-dir
/Users/chris/ws/miniforge.ai/miniforge/.git/worktrees/repo
/Users/chris/ws/miniforge.ai/miniforge/.git
```

A linked worktree owns HEAD, the index, and the working tree. `.git/config` and every other ref live in the shared common dir. So the bench's deliberate `git remote set-url origin <mirror>` — intended to stop a completed run opening a real PR — rewrote the launching checkout's origin, and its `git fetch origin main` rewrote the launching checkout's remote-tracking ref. Same class as the known workflow-runtime sandbox leak: an isolation mechanism that isolates less than the caller assumes.

## Changes

- `bb bench:provision` — clones the launching checkout, inits the bare mirror, redirects only the clone's origin. Captures the launching checkout's `remote.origin.url` before the work and re-checks after; a mismatch fails the provision. Refuses an existing `<root>/repo`.
- `bb bench:verify` — fail-closed gate comparing the bench's per-worktree git dir against its common dir, and its common dir against the launching checkout's.
- `qualified-branch-ref` — normalizes push destinations to exactly one `refs/heads/` prefix; the mirror is seeded by explicit push rather than by mirroring every ref.
- `DOGFOODING.md` §Bench Runs — the correct procedure and the failure mode.

## Correction on `refs/heads/heads/main`

Reported as a suspected refspec bug in the run's push path. It is not. The ref lives in the live checkout and predates the bench:

```
93c56dafa refs/heads/heads/main@{2026-05-19 19:31:14 -0700}: branch: Created from origin/main
```

Nine reflog entries follow over two days, all `pull --ff-only` and `reset: moving to origin/main` — a branch literally named `heads/main` that someone checked out and worked on. No committed code constructs a branch name by stripping only `refs/`, and no source file runs `git pull --ff-only`. The bench mirror acquired it because it was cloned from the bench worktree, which shares `refs/heads` with the live checkout. The generic fix (qualified refspecs, explicit-push seeding) is in this PR; deleting the stray ref is left to the operator because a bench run was in flight.

## Testing

`development/test/bench_test.clj` runs real git against throwaway repos — the leak is a property of how git shares a common dir, so a mocked git cannot observe it. The last test creates a linked worktree, runs `set-url` inside it, and shows the launching checkout's origin change.

Verified against the real leaky bench (`bb bench:verify` exits 1) and a freshly provisioned one (exits 0). Pre-commit suite green on all five commits.

### DIFF
--- a/DOGFOODING.md
+++ b/DOGFOODING.md
@@ -81,6 +81,39 @@ bb miniforge run <spec-path>
 When `gh` is authenticated but `GITHUB_TOKEN` is not exported, the wrapper
 passes `GITHUB_TOKEN=$(gh auth token)` into the Miniforge process.
 
+## Bench Runs
+
+A bench (an A/B experiment over repeated dogfood runs) points `origin` at a
+throwaway local mirror so a completed run cannot open a real PR. Provision
+the sandbox with:
+
+```bash
+bb bench:provision /path/to/launching/checkout
+```
+
+This clones to `~/.miniforge/bench/repo`, creates the bare mirror
+`~/.miniforge/bench/origin.git`, and redirects only the clone's `origin`.
+Override the root with `MINIFORGE_BENCH_ROOT`.
+
+Do **not** use `git worktree add` for a bench sandbox. A linked worktree
+owns HEAD, the index, and the working tree — `.git/config` and every other
+ref live in the shared common dir. Inside a linked worktree,
+`git remote set-url origin` rewrites the *launching* checkout's origin and
+`git fetch origin main` rewrites the *launching* checkout's
+`refs/remotes/origin/main`. On 2026-08-06 that left the live checkout
+pointing at the bench mirror with `origin/main` force-updated backwards to
+the mirror's stale head; branches cut from `origin/main` afterwards were
+based on a stale commit and pushed to the mirror instead of GitHub.
+
+A bench runner should gate on the guard before it touches anything:
+
+```bash
+bb bench:verify ~/.miniforge/bench/repo /path/to/launching/checkout
+```
+
+It exits non-zero when the bench shares a git common dir with the
+launching checkout.
+
 ## Resume, Do Not Restart
 
 When a dogfood run stops, do not re-run the same spec from scratch if a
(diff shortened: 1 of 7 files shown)


H31MDALLR:
All four review comments were real defects. Fixed in 62c35d45f.

**Eager `steps` vector** — correct, and the worst of the four. Every `(git ...)` call in the vector literal ran before any exit code was inspected, so a failed clone still init'd the mirror and attempted the push. The `failed at <step>` report named the first failure while the rest had already run. Steps are thunks now, short-circuited via `reduce`/`reduced` on the first non-zero exit. Pinned by a new test that provisions at a nonexistent ref and asserts no mirror is left behind.

**Stale mirror on re-provision** — correct. `git init --bare` exits 0 on an existing repository, so a leftover `<root>/origin.git` kept its refs and defeated the explicit-push seeding guarantee. The existence check now covers the mirror as well as the clone, in one clause.

**`absolute-path` not total** — correct. `getCanonicalPath` resolves symlinks and can throw on IO or security failures, which would have crashed `bb bench:verify` on exactly the operator-typed path the namespace exists to survive. Falls back to the uncanonicalized absolute path.

**Docstring / stratum drift** — correct. The stratum linter's autofix had renumbered `verify` to 1 after the docstring was written, and there are two entry points, not one. Reworded to name both.

6 tests / 21 assertions green, clj-kondo clean, pre-commit suite green.

PR #1686, opened 2026-08-07
docs(architecture): RFC — environment provenance at the workbench boundary

H31MDALLR:
## What this is

An RFC, status **Proposed**. No code changes — it asks for four decisions to be ratified before implementation.

[docs/architecture/rfc-workbench-environment-provenance.md](docs/architecture/rfc-workbench-environment-provenance.md)

## The gap

The workbench contract pins the *workload* an experiment arm ran — `source_hashes`, with the ETL adapter failing closed when a candidate's hash does not match its baseline's. It does not pin the *environment*. That is whatever the operator set up by hand, and no consumer of a snapshot can tell a clean arm from a contaminated one.

## The evidence

The 2026-08-06 seeded-trap bench. Beyond the git-hygiene defect fixed in #1685:

- Both arms shared task-worktree storage. The runner records it and works around it: *"Task worktrees pool here regardless of `MINIFORGE_HOME` (verified 2026-08-06) — attribute per run by before/after diff."* An attribution heuristic standing in for isolation.
- Both arms shared `refs/heads/task-*` with the live checkout, for the same reason the config leaked.

The completed baseline matrix reported a NULL primary. A null from arms that shared state is not the same evidence as a null from arms that did not, and today's contract cannot distinguish them.

## Decisions to ratify

1. `RunEnvironment` becomes a field on `WorkbenchSnapshot`.
2. The adapter **computes** `isolated`; it does not accept it. A runner that could be trusted to be isolated would not have produced the incident.
3. A non-isolated arm **fails** the resolved-run boundary — a gate, not a warning field.
4. Provisioning stays out of the adapters. The adapter gates; `bb bench:provision` provisions.

On (4): the adapter's contract is defined *after* a run reaches terminal status; provisioning happens *before*. One component with both lifecycles gets two failure modes and two sets of callers. The RFC records this and three other rejected alternatives with reasoning.

## Shape of it

The gate is one more clause in the `cond` of fail-closed guards `project` already has, beside `unresolved-run-message` and `source-hash-mismatch`. No new failure mode, no new boundary. Provision-time attestation, projection-time re-verification, missing attestation reads as `isolated false`.

Cross-arm independence — the part that actually closes the validity hole — extends the existing baseline-vs-candidate comparison from `source_hashes` to `state_root_digest`, rather than adding a new seam.

## What it does not do

Does not fix the worktree executor (separate, in flight). Does not detect contamination outside git and the state root — shared caches, ordering effects — and says so rather than overselling the field name. Does not repair completed benches; it invalidates them, which is why the gate ships alone.

## Open questions in the RFC

Container runs may already be trivially isolated via the OCI path — unverified, needs someone who knows it. Whether a pre-registered amendment can override the gate is left to whoever owns the bench methodology. `workbench-contract` is consumed by Thesium too; that needs confirming before tagging.

## Adoption order

Steps 1-3 are non-breaking and can land in any order. Step 4 is the breaking one and should be the only thing in its release.


H31MDALLR:
The three open questions are answered from source in `6dc6806fa`. None of them stays open.

**Container runs are isolated, but not trivially.** `clone-and-checkout!` (`dag-executor/executor.clj:330`) runs `git clone <repo-url> .` *inside* the environment via `execute!`, so an OCI-executed run's repository is a fresh clone with its own config and refs. The qualifier is host mounts — a bind-mounted workdir is back to sharing — and that question is already decided by existing code: `plan-security/scan-plan` reports a host runtime-socket mount as `:hard-stop` and any host mount outside the allowlist as `:review`. So `kind "container"` yields `isolated true` when the workdir is not host-mounted, and the adapter reads a verdict that already exists. No new mechanism.

**Overrides follow the house pattern.** `MINIFORGE_COMMIT_BUDGET_OVERRIDE` and `MINIFORGE_PR_BUDGET_OVERRIDE` both allow the override, require a rationale, and echo it into the trail — the same discipline the bench runsheet states for itself. Adopted, with one change: the rationale is written into the snapshot rather than a log, so a snapshot admitted over a failed environment check carries the reason with it permanently instead of in a separable log line.

**Nothing in Thesium breaks.** Verified on four points rather than assumed: no `deny_unknown_fields` anywhere in the Rust crates, and optional fields already use `Option` + `skip_serializing_if`; there is no Swift binding at all (`bindings/swift/` is a README, so adoption step 1 has been corrected to drop it); every Thesium consumer goes through `workbench-adapter-kit` validating against the contract's own `schema/WorkbenchSnapshot`, and no consumer re-declares the map or sets `:closed true`; and nothing in-tree verifies a snapshot signature, so an added field cannot invalidate a digest check that does not exist.

One incidental defect surfaced while answering the first question, recorded in the RFC for its own issue rather than fixed here: `infer-repo-url` (`cli/workflow_runner/sandbox.clj:63`) falls back to the launching directory's `git remote get-url origin` and hands that to a clone executed *inside* the container. When origin is a local path — exactly what the 2026-08-06 incident produced — the container is told to clone a host path it cannot see.

PR #1692, opened 2026-08-07
feat(dag-executor): observe the git state a task worktree shares with its host

H31MDALLR:
A linked `git worktree` owns HEAD, the index, and the working tree. `.git/config` and every ref outside HEAD live in the shared common dir. Verified against real git rather than taken from documentation:

```text
$ git -C host worktree add --detach wt
$ git -C wt remote set-url origin https://example.invalid/EVIL.git
$ git -C host config --get remote.origin.url
https://example.invalid/EVIL.git

$ git -C wt update-ref refs/remotes/origin/main HEAD
$ git -C wt branch leaked-branch
$ git -C host show-ref
... refs/remotes/origin/main   (moved)
... refs/heads/leaked-branch   (created)
```

On 2026-08-06 this reached the live miniforge checkout through the bench sandbox: `remote.origin.url` repointed at a local bare mirror, `refs/remotes/origin/main` force-updated backwards. Branches cut from `origin/main` afterwards were based on a stale commit and pushed to the mirror, so the PRs never appeared on GitHub. #1685 fixed the bench; its follow-ups named this exposure.

## Stack

1. #1691 — stratum-lint headings for `executor.clj` (merged)
2. **this PR** — the guard: snapshot and compare, no policy
3. #1693 — wrap the worktree executor, report drift

## What the guard watches

**Redirect config is the invariant.** `:clean?` is true when every key matching `^(remote\..*url|url\..*insteadof)$` came through the run unchanged.

- `remote\..*url` is anchored on `url$` rather than `\.url$`, so `pushurl` — same redirect power, less visibility — is not missed.
- `url\..*insteadof` covers `url.<base>.insteadOf` and `pushInsteadOf`. One `git config url."https://evil/".insteadOf https://github.com/` inside a task worktree redirects every subsequent fetch and push on the host without touching any `remote.*` key at all. Added after review caught it as a direct bypass.
- Values are compared as **vectors**. Git config is multi-valued: a second `remote.origin.url` line is legal and `git remote get-url` resolves to the *first*, so collapsing to one value missed a rewrite of exactly the value git uses.
- **Credentials are stripped as values are read.** No snapshot and no report holds one. The likeliest drift is the token-bearing URL the release path writes; a guard that published it while reporting it would be its own leak.

**Remote-tracking refs are reported, not enforced.** They are listed in `:ref-rewinds` and deliberately do not feed `:clean?`. A moved ref that is not a fast-forward is indistinguishable from an ordinary `git fetch` after upstream force-pushed — the default refspec is `+refs/heads/*:refs/remotes/origin/*` and the `+` forces every opportunistic update. Miniforge force-pushes its own task branches, and `release-executor/sandbox.clj:370` fetches a *parent task's branch* in the stacked-DAG shape, so enforcing here would fail ordinary `bb dogfood` runs. This is a correction from review; the first draft treated a rewind as drift. Telling a local rewrite from an upstream one needs `git ls-remote`, which this does not ask for — a follow-up.

Refs that appear or disappear are not reported at all — `fetch.prune` legitimately retires a remote-tracking ref when its upstream branch is deleted.

`drift` reports and does not judge. Fail-or-not is the caller's policy and lives in #1693.

## Testing

Real git against throwaway repos — the leak is a property of how git shares a common dir, so a mocked git cannot observe it.

- A linked worktree runs `git remote set-url origin`; the host's URL changes and `drift` reports it. The 2026-08-06 shape.
- A linked worktree sets `url.<base>.insteadOf`; drift is reported though no `remote.*` key moved.
- A **real** `git fetch` after a real fast-forward is clean.
- A **real** `git fetch` after the upstream force-pushed is also clean, with the forced update in `:ref-rewinds` as context. Driven through actual `git push --force` / `git fetch`, because the forced-update semantics only exist in the real refspec. The first version of that fixture amended an empty commit without `--allow-empty`, which fails quietly and left the branch unmoved — the test passed for no reason until that was caught.
- A ref retired by `fetch.prune` is not reported.
- A multi-valued `remote.origin.url` whose *first* value is rewritten is reported.
- A token-bearing URL is redacted; an scp-style `git@github.com:o/r.git` is left intact, since that `git@` is a username and not a secret.
- A snapshot of a path that is not a repository is an error, not an empty snapshot that would read as "nothing to compare".

Fixture remotes are `.invalid`, a reserved TLD that can never resolve, so a test that accidentally reached the network fails rather than touching a real remote.

Component suite: 427 tests, 1755 assertions, green. `bb poly:check` and `clj-kondo` clean.

## What this does NOT do

Nothing calls it yet — that is #1693, split out so the observation and the policy can be reviewed apart.

It detects; it does not prevent. Only a sandbox that does not share a common dir prevents. #1693's doc carries the full non-coverage list and follow-ups.

Doc: `docs/pull-requests/2026-08-06-host-git-guard.md`

### DIFF
--- a/components/cursor-store/src/ai/miniforge/cursor_store/store.clj
+++ b/components/cursor-store/src/ai/miniforge/cursor_store/store.clj
@@ -95,9 +95,9 @@
 
 ;; File access, and the public I/O that shares it.
 (defn- ^{:stratum 1} existing-cursors
-  "Cursors already persisted for this pipeline, or nil when no file
-   exists yet. Both callers turn a throw from here into a
-   schema/failure.
+  "Return a schema result containing cursors already persisted for this
+   pipeline, or nil cursors when no file exists yet. Malformed persisted
+   content stays on the value path used by both public operations.
 
    A file that exists but does not hold a map is a failure, not an
    empty one. `edn/read-string` returns nil for an empty file, and a
@@ -106,12 +106,14 @@
    run a success, which is the silent full re-ingest this store exists
    to prevent."
   [^java.io.File file]
-  (when (.exists file)
+  (if-not (.exists file)
+    (schema/success :cursors nil)
     (let [value (read-edn (slurp file))]
       (if (map? value)
-        value
-        (throw (ex-info (msg/t :store/not-a-map {:path (str file)})
-                        {:path (str file) :type (type value)}))))))
+        (schema/success :cursors value)
+        (schema/failure :cursors
+                        (msg/t :store/not-a-map {:path (str file)})
+                        {:path (str file) :type (type value)})))))
 
 ;------------------------------------------------------------------------------ Layer 2
 
@@ -134,23 +136,26 @@
   (try
     (let [normalized (normalize-for-storage logger cursor-map)
           file       (cursor-file pipeline-path)
-          persisted  (merge (existing-cursors file) normalized)]
-      (if (empty? normalized)
-        (do (when logger
-              (log/info logger :cursor-store :cursor-store/no-cursors
-                        {:message (msg/t :store/no-cursors)}))
-            (schema/success :cursors persisted))
-        (let [dir (.getParentFile file)]
-          (.mkdirs dir)
-          (spit file (write-edn persisted))
-          (when logger
-            (log/info logger :cursor-store :cursor-store/saved
-                      {:message (msg/t :store/saved {:count     (count normalized)
-                                                      :persisted (count persisted)})
-                       :data {:count     (count normalized)
-                              :persisted (count persisted)
-                              :path      (str file)}}))
-          (schema/success :cursors persisted))))
+          persisted-result (existing-cursors file)]
+      (if (schema/failed? persisted-result)
+        persisted-result
+        (let [persisted (merge (or (:cursors persisted-result) {}) normalized)]
+          (if (empty? normalized)
+            (do (when logger
+                  (log/info logger :cursor-store :cursor-store/no-cursors
+                            {:message (msg/t :store/no-cursors)}))
+                (schema/success :cursors persisted))
+            (let [dir (.getParentFile file)]
+              (.mkdirs dir)
+              (spit file (write-edn persisted))
+              (when logger
+                (log/info logger :cursor-store :cursor-store/saved
+                          {:message (msg/t :store/saved {:count     (count normalized)
+                                                          :persisted (count persisted)})
+                           :data {:count     (count normalized)
+                                  :persisted (count persisted)
+                                  :path      (str file)}}))
+              (schema/success :cursors persisted))))))
     (catch Exception e
       (when logger
         (log/error logger :cursor-store :cursor-store/write-failed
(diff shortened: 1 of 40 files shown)


H31MDALLR:
## Review round 1 — three corrections pushed

An adversarial pass over this branch found three real defects. All fixed here; the PR doc records each.

**1. A moved remote-tracking ref is no longer a verdict** (`fix(dag-executor): report a moved remote-tracking ref, do not enforce on it`)

The premise "a plain `git fetch` advancing `origin/main` stays clean" only holds while upstream never rewrites. The default refspec is `+refs/heads/*:refs/remotes/origin/*` — the `+` forces every opportunistic update — so an upstream force-push rewinds the local tracking ref, and the guard could not tell that from a task rewriting it.

Not an edge case here: miniforge force-pushes its own task branches, and `release-executor/sandbox.clj:370` fetches a *parent task's branch* in the stacked-DAG shape. Verified against real repos:

```text
+ fe4af2f...6c3e882 feature -> origin/feature  (forced update)
```

So `:clean?` now tracks the redirect config alone; ref movement stays in `:ref-rewinds` as context. Two new tests drive real `git push --force` / `git fetch` against a real bare upstream rather than simulating with `update-ref`.

The first version of that fixture amended an empty commit without `--allow-empty`, which fails quietly and left the branch unmoved — the test passed for no reason until that was caught.

**2. `url.<base>.insteadOf` bypassed the invariant entirely** (`fix(dag-executor): watch insteadOf rewrites, not just remote URLs`)

One `git config url."https://evil/".insteadOf https://github.com/` inside a task worktree redirects every subsequent fetch and push on the host without touching any `remote.*.url`. The watched pattern widens to `^(remote\..*url|url\..*insteadof)$`, and the vocabulary follows it — `redirect-config`, `:redirects`, `:redirect-drift`.

**3. Multi-valued config was collapsed, and credentials were published**

- `git config --get-regexp` prints one line per value and the parser kept the last, but `git remote get-url` resolves to the *first* — so rewriting the first value of a two-URL remote came back clean. Values are compared as vectors now.
- The likeliest drift is `push-with-https-fallback!` setting `remote.origin.url` to `https://x-access-token:<TOKEN>@github.com/...` and dying before it restores — and the report went to stderr, into a process-lifetime atom, and into the error payload. Credentials are stripped as values are read, so no snapshot and no report holds one.

Component suite: 427 tests, 1755 assertions, green.

H31MDALLR:
Copilot round addressed — all findings were correct.

**#1692**

- **Empty config value dropped.** Real, and the cause was subtler than the comment: the global `str/trim` on the whole `--get-regexp` output. Git prints an empty value as `remote.b.url ` with the separator intact (verified on 2.48.1), and trimming the text turned the last such line back into a line with no separator. The trim is gone; per-field trimming stays, which is what keeps a CRLF ending out of a value. New test: blanking a remote URL from inside a task worktree is drift, `:after ["" ]`, not silently absent.
- **`InterruptedException` swallowed.** Caught separately with the interrupt flag restored, matching `runtime/oci_cli.clj:215`. Exit stays 1 rather than the adapter's 130 — any non-zero reads as failure here, and naming a 130 constant would push `git` a layer up and the file over SL003. The cancellation travels on the flag, not in the returned map; the docstring says so.
- **Dangling `protocols.impl.host-guarded` pointer.** Removed. That namespace arrives in #1693, so in this PR alone the reference went nowhere.

**#1693**

- **Both stale "refused" docstrings** corrected — the wrapper's namespace docstring and the test's `second-task-id`. Both predated the change to warn-and-acquire, and a stale docstring on the policy itself is the one most likely to mislead a later change or an on-call session.

Component suite: 428 tests, 1757 assertions, green.

PR #1693, opened 2026-08-07
fix(dag-executor): fail a run whose task worktree mutated the host checkout

H31MDALLR:
Wraps the worktree executor so a task run that repoints where the host checkout pushes and fetches fails its release, announces it, and leaves a verdict that keeps announcing.

## Stack

1. #1691 — stratum-lint headings for `executor.clj` (merged)
2. #1692 — the host-git guard (base of this PR)
3. **this PR** — the wrapper, the registry wiring, and the policy

## The exposure is not latent

Two paths already run remote-mutating git inside a task worktree on the host:

- `release-executor/sandbox.clj:533` — `create-branch!` issues `git fetch origin <default-branch>` through the DAG executor. In local mode that is the worktree executor, whose `execute!` runs `sh -c` in the task worktree, so the fetch writes the host's `refs/remotes/origin/*`.
- `release-executor/sandbox.clj:244` and `git.clj:403` — the HTTPS-token push fallback does `git remote set-url origin <url-with-token>`, pushes, then restores. Same routing. A process that dies between set and restore leaves **a GitHub token persisted in the host checkout's `.git/config`**.

`release-executor/core.clj` picks sandbox mode whenever `:executor` and `:environment-id` are both present — exactly the `bb dogfood` shape. Beyond those two, an agent runs arbitrary git in its task worktree.

## `extensions.worktreeConfig` was rejected, and it does not work

```text
$ git -C wt config extensions.worktreeConfig true     # writes the HOST config
$ git -C wt config --worktree remote.origin.url https://example.invalid/PERWT.git
$ git -C wt remote set-url origin https://example.invalid/EVIL2.git
$ git config -f host/.git/config --get remote.origin.url
https://example.invalid/EVIL2.git                     # host still corrupted
$ git -C wt config --get remote.origin.url
https://example.invalid/PERWT.git                     # sandbox reads clean
```

A plain `git remote set-url` — what an agent actually types — still writes the shared config. All the override buys is that the sandbox no longer *sees* the damage it did. It fixes no refs, and enabling it is itself a write to the host's config.

## What happens on drift, and what does not

`release-environment!` returns `:host-git-drift`, warns on stderr, and records a verdict against the checkout. Every later acquire on it warns again for the life of the process. `reset-guard-state!` clears it once the checkout is repaired.

The stderr warning is not decoration: every current caller of `release-environment!` discards its result — `with-environment` calls it in a `finally`, `release-execution-environment!` wraps it in a bare `try` — so a returned error alone would have been as silent as the leak.

**A condemned checkout is warned about, not refused, and that is a correction from review.** The first version refused it. In `:local` mode that would have been strictly worse than the drift: `workflow/runner_environment.clj:126` collapses any non-ok acquire to `nil`, and `workflow/runner.clj:319` then runs the pipeline with the caller's original opts — no executor, no `:execution/worktree-path`. That is the state `runner_environment`'s own comment describes as what "let a caller that drove run-pipeline without a sandbox leak a real commit into the live worktree". Making a refusal actually stop the run needs `build-env-record` to tell "no executor configured" from "acquire failed" — a workflow-component change, itself blocked behind a namespace split there. Follow-up 1. **This PR does not claim fail-closed enforcement it does not have.**

Verdicts are keyed on the canonicalized `git rev-parse --git-common-dir`, so `"."`, an absolute path, and a linked worktree of the same checkout are one key. `getCanonicalPath`, not `getAbsolutePath` — the latter leaves `/./` in place, which the test for it proved.

A post-release snapshot that cannot be read records a verdict and warns rather than passing the error through: the checkout became unreadable mid-run, so drift is unknown, and unknown is the one case the guard must not wave through.

Attribution is per-checkout, not per-task. With tasks in parallel against one checkout, a mutation between an acquire and its release is attributed to whichever run releases next.

`executor-type` reports the delegate's own `:worktree`, so `runner-environment/select-capsule-executor`'s governed-mode check is unaffected. Split across two namespaces because the record sits a layer above the lifecycle steps and its factory a layer above that — four bands, which rule 210 answers with a split.

## Testing

A real host checkout, a real `git worktree add`, a real config write from inside the task worktree. The whole lifecycle, not a stubbed delegate — a stub would pass with the guard removed.

- A clean run releases exactly as the unwrapped executor does. The regression that matters most: this executor is the fallback `bb dogfood` uses.
- A run that repoints origin fails its release and warns.
- The next task against the condemned checkout warns again while still being handed a worktree. Asserted with a distinct task id, so a stale branch from the first task cannot be what makes it pass.
- A verdict follows the *checkout*, not the path spelling: an acquire naming it as `<path>/./` warns too.
- Resetting clears the verdict, and a repaired checkout releases clean.

Verified the guard is load-bearing rather than assumed — same run, **unwrapped** executor:

```text
UNGUARDED release ok? => true
host origin after     => https://example.invalid/redirected-mirror.git
```

Component suite: 427 tests, 1755 assertions, green, including the pre-existing worktree tests pinning `resolve-branch-sha`, the three-attempt create fallback, and the archive/restore paths. `bb poly:check` and `clj-kondo` clean.

## What this does NOT fix

- **It does not prevent anything.** A run can still rewrite the host's config and refs; the guard notices afterwards. Repair is by hand, and a later run against the drifted checkout proceeds. Only a sandbox that does not share a common dir prevents.
- **Enforcement is announcement, not refusal** — see the policy section and follow-up 1.
- `refs/heads` is not watched — `git worktree add -b task-<id>` creates a branch in the host by design.
- Config keys other than redirect config are not watched (`core.hooksPath`, `credential.helper`).
- Detection is per-process; a verdict does not survive a restart.
- The release-executor's `set-url`/restore pair is untouched — the guard catches a failed restore after the fact, it does not stop the token being written.
- `worktree.clj` itself is unchanged (follow-up 5).

## Follow-ups

1. **Make a refused acquire actually stop the run** — `runner_environment.clj:126` must tell "no executor configured" from "acquire failed", and the latter must abort in both modes. Blocked behind a namespace split there (SL003, 5 layers).
2. **Distinguish a local ref rewrite from an upstream one** with `git ls-remote`, so a rewound tracking ref can rejoin the invariant.
3. **Stop writing tokens into git config at all** — push straight to the token-bearing URL in `release-executor/git.clj:399-439` and `sandbox.clj:227-265`. Nothing written to config, nothing to restore, nothing to leak on a mid-sequence death. Their tests pin the current pair, so it is a deliberate behaviour change.
4. **Clone-based task sandboxes** behind a flag, defaulting off. Needs `git clone --local` (hardlinked objects, near-nil marginal disk), the clone's `origin` set to the host's origin *URL* not path, the host repo path threaded into `notify-file-written!` so scratch refs survive release, and `rm -rf` in place of `git worktree remove`.
5. **`worktree.clj` cannot currently be edited.** No `Layer N` headings, so stratum-lint skips it — but staging it triggers `--fix`, which produces a 919-line rewrite and then fails SL003 at **7 distinct layers**, blocking the commit. That is why this PR wraps from the registry rather than modifying the executor.
6. **Watch more of the shared config surface** — `core.hooksPath` and `credential.helper` turn a task worktree into host code execution or credential capture.

Doc: `docs/pull-requests/2026-08-06-worktree-host-git-wiring.md`

### DIFF
--- a/components/cursor-store/src/ai/miniforge/cursor_store/store.clj
+++ b/components/cursor-store/src/ai/miniforge/cursor_store/store.clj
@@ -95,9 +95,9 @@
 
 ;; File access, and the public I/O that shares it.
 (defn- ^{:stratum 1} existing-cursors
-  "Cursors already persisted for this pipeline, or nil when no file
-   exists yet. Both callers turn a throw from here into a
-   schema/failure.
+  "Return a schema result containing cursors already persisted for this
+   pipeline, or nil cursors when no file exists yet. Malformed persisted
+   content stays on the value path used by both public operations.
 
    A file that exists but does not hold a map is a failure, not an
    empty one. `edn/read-string` returns nil for an empty file, and a
@@ -106,12 +106,14 @@
    run a success, which is the silent full re-ingest this store exists
    to prevent."
   [^java.io.File file]
-  (when (.exists file)
+  (if-not (.exists file)
+    (schema/success :cursors nil)
     (let [value (read-edn (slurp file))]
       (if (map? value)
-        value
-        (throw (ex-info (msg/t :store/not-a-map {:path (str file)})
-                        {:path (str file) :type (type value)}))))))
+        (schema/success :cursors value)
+        (schema/failure :cursors
+                        (msg/t :store/not-a-map {:path (str file)})
+                        {:path (str file) :type (type value)})))))
 
 ;------------------------------------------------------------------------------ Layer 2
 
@@ -134,23 +136,26 @@
   (try
     (let [normalized (normalize-for-storage logger cursor-map)
           file       (cursor-file pipeline-path)
-          persisted  (merge (existing-cursors file) normalized)]
-      (if (empty? normalized)
-        (do (when logger
-              (log/info logger :cursor-store :cursor-store/no-cursors
-                        {:message (msg/t :store/no-cursors)}))
-            (schema/success :cursors persisted))
-        (let [dir (.getParentFile file)]
-          (.mkdirs dir)
-          (spit file (write-edn persisted))
-          (when logger
-            (log/info logger :cursor-store :cursor-store/saved
-                      {:message (msg/t :store/saved {:count     (count normalized)
-                                                      :persisted (count persisted)})
-                       :data {:count     (count normalized)
-                              :persisted (count persisted)
-                              :path      (str file)}}))
-          (schema/success :cursors persisted))))
+          persisted-result (existing-cursors file)]
+      (if (schema/failed? persisted-result)
+        persisted-result
+        (let [persisted (merge (or (:cursors persisted-result) {}) normalized)]
+          (if (empty? normalized)
+            (do (when logger
+                  (log/info logger :cursor-store :cursor-store/no-cursors
+                            {:message (msg/t :store/no-cursors)}))
+                (schema/success :cursors persisted))
+            (let [dir (.getParentFile file)]
+              (.mkdirs dir)
+              (spit file (write-edn persisted))
+              (when logger
+                (log/info logger :cursor-store :cursor-store/saved
+                          {:message (msg/t :store/saved {:count     (count normalized)
+                                                          :persisted (count persisted)})
+                           :data {:count     (count normalized)
+                                  :persisted (count persisted)
+                                  :path      (str file)}}))
+              (schema/success :cursors persisted))))))
     (catch Exception e
       (when logger
         (log/error logger :cursor-store :cursor-store/write-failed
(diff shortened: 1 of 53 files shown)


H31MDALLR:
## Review round 1 — the enforcement claim was wrong, and is now corrected

The adversarial pass found that the acquire-time refusal **inverted its own intent** in `:local` mode. Pushed a fix; the PR doc and the body above are updated to match.

**A refused acquire did not fail the run — it ran the pipeline with no sandbox.**

`workflow/runner_environment.clj:126` (`build-env-record`) collapses any non-ok acquire to `nil`. `workflow/runner.clj:319` then proceeds with the caller's original opts — no executor, no `:execution/worktree-path`. That is the state `runner_environment`'s own comment describes as what "let a caller that drove run-pipeline without a sandbox leak a real commit into the live worktree".

So the sticky verdict was backwards: the first drifted run condemned the checkout, and every later run in that JVM would have executed directly against the host. Strictly worse than the drift it was reacting to.

A condemned checkout is now **warned about on every acquire, not refused**. Making a refusal actually stop the run needs `build-env-record` to tell "no executor configured" from "acquire failed" — a change in the workflow component, itself blocked behind a namespace split there (`runner_environment.clj` fails SL003 at 5 layers). That is now follow-up 1, and the PR no longer claims fail-closed enforcement it does not have.

Two other fail-closed gaps, also fixed:

- **A post-release snapshot that could not be read was returned silently**, while acquire treated the same condition as fatal. A checkout that became unreadable mid-run was waved through. It now records a verdict and warns — unknown is the one case the guard must not pass.
- **Verdicts were keyed on the caller's path string**, so `"."`, an absolute path, and a linked worktree of the same checkout were four keys for one shared common dir. Keyed on the canonicalized `git rev-parse --git-common-dir` now. `getCanonicalPath`, not `getAbsolutePath` — the latter leaves `/./` segments in place, which the new test proved.

Also noted in the doc rather than fixed: attribution is per-checkout, not per-task. With tasks in parallel against one checkout, a mutation between an acquire and its release is attributed to whichever run releases next. The verdict on the checkout is right; the run it names may not be the culprit.

Component suite: 427 tests, 1755 assertions, green.

H31MDALLR:
Copilot round addressed — all findings were correct.

**#1692**

- **Empty config value dropped.** Real, and the cause was subtler than the comment: the global `str/trim` on the whole `--get-regexp` output. Git prints an empty value as `remote.b.url ` with the separator intact (verified on 2.48.1), and trimming the text turned the last such line back into a line with no separator. The trim is gone; per-field trimming stays, which is what keeps a CRLF ending out of a value. New test: blanking a remote URL from inside a task worktree is drift, `:after ["" ]`, not silently absent.
- **`InterruptedException` swallowed.** Caught separately with the interrupt flag restored, matching `runtime/oci_cli.clj:215`. Exit stays 1 rather than the adapter's 130 — any non-zero reads as failure here, and naming a 130 constant would push `git` a layer up and the file over SL003. The cancellation travels on the flag, not in the returned map; the docstring says so.
- **Dangling `protocols.impl.host-guarded` pointer.** Removed. That namespace arrives in #1693, so in this PR alone the reference went nowhere.

**#1693**

- **Both stale "refused" docstrings** corrected — the wrapper's namespace docstring and the test's `second-task-id`. Both predated the change to warn-and-acquire, and a stale docstring on the policy itself is the one most likely to mislead a later change or an on-call session.

Component suite: 428 tests, 1757 assertions, green.

PR #1741, opened 2026-08-10
docs(specs): N11 task capsule isolation spec completion (0.2.0 → 0.3.0-draft)

H31MDALLR:
MINIFORGE_PR_BUDGET_OVERRIDE: Single normative spec document. A spec amendment is atomic. Well below the ~10-12k threshold where Copilot declines to review. Docs-only, no runtime code.

## Overview

Fixes a five-way section-numbering collision, marks an implementation-mapping section informative, defers secrets and evidence to their owning specs, and adds Annex A.

Full detail: [`docs/pull-requests/2026-08-10-chris-n11-task-capsule-spec-complete.md`](docs/pull-requests/2026-08-10-chris-n11-task-capsule-spec-complete.md)

## Why

**Five duplicate section numbers.** §11's subsections were numbered §10.1–§10.5, colliding with the TaskExecutor protocol's own §10.1–§10.5. Both inbound `N11 §10` references — from N11-delta-runtime-adapter — mean the protocol, so the protocol keeps its numbers and §11's were the misnumbered ones.

**A normative section pinned to file and line numbers.** §11 maps requirements to locations like `docker.clj:428–442` inside the normative body without being marked informative. Those coordinates rot: the `docker.clj` it cites no longer exists anywhere in the tree.

**Two contracts restated from their owners.** §8.1 forbade secrets being "logged or included in evidence records" — that is N3 §8's contract, inherited by N6 §7.2. §9.1's evidence keys are not N6 artifact fields, so the record cannot be produced conformantly — the same shape as N10 §12.2.

## Changes

- **§11** renumbered to §11.1–§11.5 and marked informative, with a note that its citations are a snapshot and one is already stale.
- **§8.1** defers to N3 §8, keeping the one genuinely capsule-specific rule: a secret's *name* and *scope* may be recorded, so an auditor can see which secrets a capsule held without seeing their values.
- **§9.1** gated on registering its keys in N6 §3.1.1, with the interim position stated.

## Annex A — implementation conformance status (informative)

- **Implemented** — `components/dag-executor` provides the TaskExecutor protocol with three implementations: worktree, host-guarded, Kubernetes.
- **Specified, not implemented** — no Docker executor, so `:docker` is unselectable among the runtime classes §5 admits. No capsule evidence record. No artifact export gate, so `:export/missing-required-artifact` has no producer. No network-policy enforcement.
- **Structural** — secret teardown is stated and unverified. And §9.3's prohibition on agents resolving their workspace from `System/getProperty "user.dir"` in governed mode is **unenforced** — that is the exact fallback behind the sandbox-leak defect observed in this repository during this work, which makes the prohibition load-bearing rather than theoretical.

## Validation

Documentation only; no runtime code touched.

- `markdownlint` clean on all three files
- Verified no duplicate section numbers remain
- Verified both inbound `N11 §10` references mean the TaskExecutor protocol **before** renumbering
- Verified `docker.clj` is absent from the tree and the other cited files resolve

### DIFF
--- a/docs/pull-requests/2026-08-10-chris-n11-task-capsule-spec-complete.md
+++ b/docs/pull-requests/2026-08-10-chris-n11-task-capsule-spec-complete.md
@@ -1,0 +1,99 @@
+<!--
+  Title: Miniforge.ai
+  Author: Christopher Lester (christopher@miniforge.ai)
+  Copyright 2025-2026 Christopher Lester. Licensed under Apache 2.0.
+-->
+
+# docs: N11 task capsule isolation spec completion (0.2.0 → 0.3.0-draft)
+
+## Overview
+
+Fixes a section-numbering collision, marks an implementation-mapping section
+informative, defers secrets and evidence to their owning specs, and adds
+Annex A.
+
+## Motivation
+
+**Five duplicate section numbers.** §11's subsections were numbered §10.1
+through §10.5, colliding with the TaskExecutor protocol's own §10.1–§10.5.
+Both inbound `N11 §10` references — from N11-delta-runtime-adapter — mean the
+protocol, so the protocol keeps its numbers and §11's subsections were the ones
+misnumbered.
+
+**A normative section pinned to file and line numbers.** §11 maps requirements
+to locations like `docker.clj:428–442` inside the normative body, without being
+marked informative. Those coordinates rot: the `docker.clj` it cites no longer
+exists anywhere in the tree.
+
+**Two contracts restated from their owners.** §8.1 forbade secrets being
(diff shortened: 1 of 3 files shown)


H31MDALLR:
On the 0.18.0 gap: keeping 0.19.0-draft. [#1739](https://github.com/miniforge-ai/miniforge/pull/1739) (N1) supplies 0.18.0-draft and is ahead in review, so it lands first and the sequence becomes contiguous. This branch was cut from the same main, which is why 0.18.0 is not visible here yet — the version-history block will conflict on rebase and I will resolve it then, as I did for the N6/N2 and N8/N9/N10 pairs. If N1 were to land after this one I would renumber instead.

PR #1797, opened 2026-08-17
spec: require a kustomize-readable deploy preflight basis

H31MDALLR:
## Overview

Amends `work/ariadne-deploy-grant-enforcement.spec.edn` so the queued enforcement
work cannot satisfy its acceptance criteria with a gate that approves every
manifest. Spec text only; no source or behavior change.

## The finding

`check-resource-count` and `check-gke-node-limit` parse Pulumi preview JSON.
`policy/parse-preview` calls `json/parse-string` and swallows the exception to
`{}`, so against a kustomize render there are no steps, no violations, and
`deploy-authority/preflight` answers `:allow` for any manifest.

Measured on `1eab1888b`:

```text
preflight over 5000-doc kustomize render => #:preflight{:result :allow, :violations []}
```

`deploy-governed/transact!` does call both checks before the apply. The criteria
"check-X is evaluated before the real apply" were therefore satisfiable by a gate
structurally unable to deny — evaluation recorded, no decision made. Worse than an
absent gate, because the durable effect record attests to a check that could not
have failed.

Kustomize is the primary Kubernetes provisioning path here, so the checks become
kustomize-aware rather than the deploy seam growing a separate basis.

## Changes

- Record the finding and its measurement in `:spec/description`.
- Replace the two evaluation-only criteria with denial criteria.
- Require a per-check test asserting a violation over a kustomize render.
- Constrain every preflight check to parse the artifact the configured provider
  renders; a check that cannot return a violation is not left in the path.
- Record the provider-shape mismatches found while checking whether #1713's
  operations satisfy `transact!`: `dry-run!` is 2-arity against a 1-arity call
  site, and `apply!` returns a schema-shaped map with no `:deploy/failed?`, so a
  failed apply reads falsy in `commit!` and never records
  `:effect/outcome :failed`. `:observe!` and `:rollback-info!` already match.
- Add a byte-identity criterion between the proposal's recorded manifest and the
  bytes applied — the reason #1713 added `render!` / `apply-rendered!` / `dry-run!`.

Two sub-decisions are named as belonging to the implementation, not fixed here:
whether to add a YAML reader (phase-deployment carries cheshire and malli only)
or split documents structurally; and what `check-gke-node-limit` means over a
kustomize render, given GKE node pools are infrastructure rather than manifest
resources.

## Testing

- spec-parser validates the amended file: `parse-spec-file` ok,
  `valid-spec-payload?` true, `validate-spec` `{:valid? true}`.
- `work/QUEUE.md` row derives from title, tier, theme, and axes — all unchanged.
- Pre-commit suite green (8 tests, 634 assertions).
- No source changes, so no component suite is affected.

## Related

- #1713 — exact deployment preflight operations (merged); supplies the operations
  this spec requires the enforcement work to bind correctly.
- #1752 — the deploy governance seam this spec finishes wiring.


H31MDALLR:
@copilot resolve the merge conflicts in this pull request

